PR #35, opened 2023-04-11
Fix cmake and gcc warnings

filip-stenstrom:
Fixes a number of cmake and gcc warnings.


Test/FMI2/fmi2_xml_parsing_test.c
```diff
@@ -43,20 +41,20 @@ void test_parser(char *xml_dir, int should_not_log_expected_msg, int configurati
...
+    full_path = concat(g_name_check_test_directory, xml_dir);
     fmu = fmi2_import_parse_xml(context, full_path, NULL);
     fmi_import_free_context(context);
     free(full_path);
 
     if (fmu == NULL) {
         goto err1;
     }
-    if (!should_not_log_expected_msg && did_not_log_expected_msg ||
-            did_not_log_expected_msg && !should_not_log_expected_msg) {
```
filip-stenstrom, line 57: This logic was incorrect.

src/XML/src/FMI3/fmi3_xml_dimension_impl.h
```diff
@@ -30,7 +30,7 @@ struct fmi3_xml_dimension_t {
 };
 
 /* declare fmi3_xml_dimension_t vector functions */
-jm_vector_declare_template(fmi3_xml_dimension_t)
+jm_vector_declare_template_no_typedef(fmi3_xml_dimension_t)
```
filip-stenstrom, line 33: The typedef exists in the public header file, which means that gcc will give a (very long) warning if we add another typedef here when we declare the template.

Config.cmake/fmiimport.cmake
```diff
@@ -124,7 +124,6 @@ PREFIXLIST(FMIIMPORTSOURCE  ${FMIIMPORTDIR}/)
 
 add_library(fmiimport ${FMILIBKIND} ${FMIIMPORTSOURCE} ${FMIIMPORTHEADERS})
 target_link_libraries(fmiimport ${JMUTIL_LIBRARIES} ${FMIXML_LIBRARIES} ${FMIZIP_LIBRARIES} ${FMICAPI_LIBRARIES})
-add_dependencies(fmiimport generate_functions_import)
```
filip-stenstrom, line 127: `generate_functions_import` no longer exists

Config.cmake/fmixml.cmake
```diff
@@ -227,8 +227,6 @@ ExternalProject_Add_Step(
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/ExpatEx
 )
 
-add_dependencies(expatex ${CMAKE_BINARY_DIR}/CMakeCache.txt ${FMILIBRARYHOME}/CMakeLists.txt)
```
filip-stenstrom, line 230: Can't put dependencies on files. Not certain this ever worked.

Test/FMI1/fmi1_xml_parsing_test.c
```diff
@@ -11,14 +11,14 @@ static const int NO_LOG_EXPECTED_MSG = 0;
 static const int DO_LOG_EXPECTED_MSG = 1;
 
 static int did_log_expected_msg;
-static char *expected_message = "Invalid structured ScalarVariable name";
-static char *name_check_test_directory;
+static char *g_name_check_test_directory;
+static char *g_expected_message = "Invalid structured ScalarVariable name";
```
filip-stenstrom, line 15: Not a warnings fix, but added this as well. I think we should use `s_` for static variables instead of `g_` but changed it like this for consistency for now.

src/CAPI/src/FMI1/fmi1_capi.c
```diff
@@ -35,14 +35,13 @@ extern "C" {
...
 {
     char fname[FUNCTION_NAME_LENGTH_MAX];
-    int len;
 
     if (strlen(fmu->modelIdentifier) + strlen(function_name) + 2 > FUNCTION_NAME_LENGTH_MAX) {
         jm_log_fatal(fmu->callbacks, FMI_CAPI_MODULE_NAME, "DLL function name is too long. Max name length is set to %s.", STRINGIFY(FUNCTION_NAME_LENGTH_MAX));
         return jm_status_error;
     }
 
-    len = jm_snprintf(fname,FUNCTION_NAME_LENGTH_MAX,"%s_%s",fmu->modelIdentifier, function_name);
```
filip-stenstrom, line 45: return value not used

src/XML/src/FMI1/fmi1_xml_variable.c
```diff
@@ -863,8 +863,6 @@ void fmi1_xml_eliminate_bad_alias(fmi1_xml_parser_context_t *context, jm_vector(
     size_t n, index;
 
     fmi1_xml_variable_t* v = (fmi1_xml_variable_t*)jm_vector_get_item(jm_voidp)(varByVR, indexVR);
-    fmi1_value_reference_t vr = v->vr;
-    fmi1_base_type_enu_t vt = fmi1_xml_get_variable_base_type(v);
```
filip-stenstrom, line 867: Never used

src/XML/src/FMI3/fmi3_xml_dimension.c
```diff
@@ -19,8 +19,6 @@
 
 #include "FMI3/fmi3_xml_variable.h"
 
-static const char* module = "FMI3XML";
```
filip-stenstrom, line 22: Never used

Test/FMI2/fmi2_xml_parsing_test.c
```diff
@@ -7,29 +7,27 @@
...
 
-void test_parser(char *xml_dir, int should_not_log_expected_msg, int configuration)
+void test_parser(char *xml_dir, int should_log_expected_msg, int configuration)
 {
     jm_callbacks cb;
     fmi_import_context_t *context;
     fmi2_import_t *fmu;
     char *full_path;
-    int res = 0;
+    int res = 1;
```
filip-stenstrom, line 30: Test was incorrectly always succeeding before.

PeterMeisrimelModelon: I would propose not assigning any value here at all and doing it more explicitly based on the result of the test.

PeterMeisrimelModelon (approved):

Test/FMI1/fmi1_xml_parsing_test.c
filip-stenstrom: Well they are not truly global, they are still local to the file/compilation unit. I meant we'd then distinguish between:

```
char* g_str = ...;
static char* s_str = ...;
```

I mean, FMIL has no global variable, only static ones, so would prefer to call them for what they are.

Test/FMI2/fmi2_xml_parsing_test.c
filip-stenstrom: Yes, the test is a bit convoluted. It seems to try to handle deallocation even on test failure, making things complicated. Moving testing to catch will also make these things easier.

PR #40, opened 2023-04-24
Parsing, API and tests for array variables

modelonrobinandersson:
- [X] float64 - was already implemented
- [X] float32 - was already implemented
- [X] int64
- [X] int32
- [X] int16
- [X] int8
- [X] uint64
- [X] uint32
- [X] uint16
- [X] uint8
- [X] string
- [X] bool
- [X] binary
- [X] enum


src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -640,6 +640,86 @@ gen_fmi3_xml_set_attr_TYPEXX(uint32,  intXX)
         jm_log_error(context->callbacks, module, "Unable to allocate memory. File: %s, Line: %d", __FILE__, __LINE__);      \
     } while (0);
 
+/**
+ * Parses a string to an array of int values
+*/
+static int fmi3_xml_str_to_array_intXX(
```
modelonrobinandersson, line 646: I will refactor this method next to remove code duplication between all the types.

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -743,13 +823,17 @@ int fmi3_xml_set_attr_array(fmi3_xml_parser_context_t *context, fmi3_xml_elm_enu
     }
 
     /* write all attributes to array of correct type */
-    if (fmi3_xml_str_to_array_floatXX(context, str, arrPtr, arrSize, primType)) {
+    if (fmi3_base_type_enu_is_float(primType->baseType) && fmi3_xml_str_to_array_floatXX(context, str, arrPtr, arrSize, primType)) {
```
filip-stenstrom: I suggest you write it like:

```
if (fmi3_base_type_enu_is_float(primType->baseType)) {
   if (fmi3_xml_str_to_array_floatXX(...)) {
        ...
   }
} else if { ...
```
Otherwise you'll also try to parse as intXX even if it succeeded for floatXX. I mean it try to eval the else-if branch.

src/XML/src/FMI3/fmi3_xml_parser.c
modelonrobinandersson: This was done in commit #03e4e6a

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -30,18 +30,49 @@ static const char* module = "FMI3XML";
...
-static struct fmi3_xml_variable_default_values VARIABLE_DEFAULT_VALUES = { 0, 0 };
+static struct fmi3_xml_variable_default_values VARIABLE_DEFAULT_VALUES = \
+    { 0, 0, 0, 0, 0, 0, 0, 0, 0, "" };
 
-static void* fmi3_xml_get_type_default_value(fmi3_bitness_enu_t bitness) {
-    switch (bitness) {
-        case fmi3_bitness_32:
-            return &VARIABLE_DEFAULT_VALUES.float32;
-        case fmi3_bitness_64:
+static void* fmi3_xml_get_type_default_value(fmi3_base_type_enu_t baseType) {
```
filip-stenstrom, line 48: This impl feels like it's too much overhead now. Easier to just use add a field to each baseType and then cast to void*. I mean the whole point was to avoid needing to cast, but it wasn't really a good idea from the start.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1129,15 +1173,17 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
...
-        fmi3_xml_model_description_t* md = context->modelDescription;
-        fmi3_xml_type_definitions_t* td = &md->typeDefinitions;
-        fmi3_xml_variable_t* variable = jm_vector_get_last(jm_named_ptr)(&md->variablesByName).ptr;
-        fmi3_xml_variable_type_base_t * declaredType = 0;
-        fmi3_xml_int_type_props_t * type = 0;
-        int hasStart;
+    md = context->modelDescription;
+    td = &md->typeDefinitions;
+    variable = jm_vector_get_last(jm_named_ptr)(&md->variablesByName).ptr;
+    type = 0;
```
filip-stenstrom, line 1179: Unintentional line?

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1129,15 +1173,17 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
...
 
-        declaredType = fmi3_get_declared_type(context, elmID, &defaultType->super) ;
+    if(!data) {
+        fmi3_xml_variable_type_base_t* declaredType = fmi3_get_declared_type(
+            context,
+            elmID,
+            &defaultType->super
+        );
```
filip-stenstrom, line 1186: I suggest we don't create a new line for each argument, but just do double indentation on next line. That's how we do it almost everywhere else. Unless it greatly increases readability that is.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1168,36 +1214,52 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
...
         if (hasStart) {
-            fmi3_xml_variable_start_int_t * start = (fmi3_xml_variable_start_int_t*)fmi3_xml_alloc_variable_type_start(td, &type->super, sizeof(fmi3_xml_variable_start_int_t));
-            fmi3_int_union_t defVal;
+            jm_string startAttr = variable->startAttr;
+
+            fmi3_xml_variable_start_int_t* start = \
+                    (fmi3_xml_variable_start_int_t*)fmi3_xml_alloc_variable_type_start(
+                        td,
+                        variable->type,
+                        sizeof(fmi3_xml_variable_start_int_t));
```
filip-stenstrom, line 1236:
```suggestion
            fmi3_xml_variable_start_int_t* start;
            start = (fmi3_xml_variable_start_int_t*)fmi3_xml_alloc_variable_type_start(td, variable->type,
                        sizeof(fmi3_xml_variable_start_int_t));
```
At least that's how I'd format it.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1168,36 +1214,52 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
...
             if (!start) {
-                fmi3_xml_parse_fatal(context, "Could not allocate memory");
+                fmi3_xml_parse_fatal(context, "Could not allocate memory for an integer start array");
```
filip-stenstrom, line 1238: I don't think we should add specific message to memory allocation failure. It doesn't really matter where it happens if it happens, and it's easier to just mentally ignore this kind of code if it's the same everywhere.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1168,36 +1214,52 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
...
                 return -1;
             }
+            if (fmi3_xml_variable_is_array(variable)) {
+                size_t nArr; /* TODO: do something with this, e.g. dimension size verification */
+                if (fmi3_xml_set_attr_array(
+                        context, elmID, fmi_attr_id_start, 0,
+                        (void**)&start->start, &nArr, startAttr, primType)) {
```
filip-stenstrom, line 1245:
```suggestion
                if (fmi3_xml_set_attr_array(context, elmID, fmi_attr_id_start, 0,
                        (void**)&start->start, &nArr, startAttr, primType))
                {
```

filip-stenstrom (commented):
For the work so far.

Also: Can you double check that the memory gets deallocated for intXX arrays? Memcheck should catch it but still.

src/XML/src/FMI3/fmi3_xml_variable.c
modelonrobinandersson: See if I understood you correctly, it is commit 3e291fc.

src/XML/src/FMI3/fmi3_xml_variable.c
modelonrobinandersson: Good catch, see 290c292.

src/XML/src/FMI3/fmi3_xml_variable.c
modelonrobinandersson: Fixed in 7465def

src/XML/src/FMI3/fmi3_xml_variable.c
modelonrobinandersson: I agree but I liked the idea that you dont have to attach a debugger in order to find where the error happens (when it happens). Do you prefer that I remove the recently added piece?

src/XML/src/FMI3/fmi3_xml_variable.c
modelonrobinandersson: Good suggestion I have updated it. See 7912e34.

src/XML/src/FMI3/fmi3_xml_variable.c
modelonrobinandersson: See 1d056fa

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1168,36 +1214,52 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
...
-               - here I'm just giving a valid value for the type so it's defined on our side at least.
-            */
-            defVal.scalar64s = 0; /* set the whole bitfield to 0 - this will evaluate to '0' for all intXX types */
-
-            if (fmi3_xml_set_attr_intXX(context, elmID, fmi_attr_id_start, 0, &start->start, &defVal, primType)) {
-                jm_log_error(context->callbacks, module, "failed to parse start value for integer");
-                return -1;
+                if (fmi3_xml_set_attr_intXX(
+                        context, elmID, fmi_attr_id_start,0,
+                        &start->start, fmi3_xml_get_type_default_value(primType->baseType), primType)) {
```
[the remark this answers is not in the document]
modelonrobinandersson: Let us discuss this tomorrow, I will leave this thread open.

src/XML/src/FMI3/fmi3_xml_parser.c
modelonrobinandersson: Let us also discuss this piece tomorrow.

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: Yes, for consistency. And my point is, if you get out-of-memory, it typically won't be deterministic where it happens, so the details of where it happens don't really matter.

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: There was also a { on a new line, but OK!

Test/FMI3/parser_test_xmls/arrays/valid/modelDescription.xml
```diff
@@ -74,12 +74,22 @@
       <Dimension start="2"/>
     </Float32>
 
+    <Int64 name="dummy_int64_array" valueReference="109" causality="input" description="Array used to test int64 get/set" start="54 23 11 5123">
+      <Dimension start="2"/>
+      <Dimension start="2"/>
+    </Int64>
+
+    <Int64 name="dummy_int64_array2" valueReference="110" causality="input" description="Array used to test the largest and smalles value of int64" start="9223372036854775807 9223372036854775806 9223372036854775805 -9223372036854775808 0 -9223372036854775807">
```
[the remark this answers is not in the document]
filip-stenstrom: In my latest ticket I tried to keep them such that I could have split-screen in VS code and see both. That ended up being like 90 lines. Might be a bit restrictive. Try to not go over 120 cols as a guideline I suppose.

I ended up with something like this:

```
<Binary name="binaryAllAttr" valueReference="4" variability="discrete"
        causality="output" initial="exact" description="myDesc"
        previous="4001" clocks="4002" mimeType="myMimeType" maxSize="444"/>

```

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: No, let's discuss it today

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -664,51 +668,63 @@ static int fmi3_xml_str_to_array_floatXX(fmi3_xml_parser_context_t* context, con
...
     assert(str);
 
-    /* create a copy that it's OK that strtok mutates */
+    /* Check first that the current function has implemented support for the type passed to this function.
+    *   Otherwise we will have errors further down where check if the parsing of scalars/arrays is OK.
+    */
+    if (!(fmi3_base_type_enu_is_int(primType->baseType) ||
+          fmi3_base_type_enu_is_float(primType->baseType))) {
+        assert(0);
+    }
```
filip-stenstrom, line 677: No need for an `if`, just place it directly in the assert

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -664,51 +668,63 @@ static int fmi3_xml_str_to_array_floatXX(fmi3_xml_parser_context_t* context, con
...
+
+    /* Create a copy that it's OK that strtok mutates */
     strCopy = context->callbacks->malloc(strlen(str) + 1); /* plus one for null character */
     if (!strCopy) {
         JM_LOG_ERROR_NO_MEM();
         res = -1;
-        goto err1;
+        goto strCopy_malloc_error;
```
filip-stenstrom, line 684: Please keep names `err1`, and only add new labels for cleanup i.e. freeing memory. It's kind of an "error stack", where you "push" a new label every time you allocate memory that needs to be freed on (local) failure.
See:
https://stackoverflow.com/questions/788903/valid-use-of-goto-for-error-management-in-c

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -664,51 +668,63 @@ static int fmi3_xml_str_to_array_floatXX(fmi3_xml_parser_context_t* context, con
...
     if (nVals == 1) { /* scalar */
-        if (fmi3_xml_str_to_floatXX(context, 1, vals, NULL, str, primType)) {
-            jm_log_error(context->callbacks, module, "Unable to parse to %s: %s", primType->name, str);
-            res = -1;
-            goto err2;
+        if ((fmi3_base_type_enu_is_int(primType->baseType) &&
+                fmi3_xml_str_to_intXX(context, 1, vals, NULL, str, primType))
+                ||
+            (fmi3_base_type_enu_is_float(primType->baseType) &&
+                fmi3_xml_str_to_floatXX(context, 1, vals, NULL, str, primType))) {
```
filip-stenstrom, line 702: We were going to discuss this on another place (and I think this is hard to read - you can break down large expressions to local variables if that increases readability)

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -664,51 +668,63 @@ static int fmi3_xml_str_to_array_floatXX(fmi3_xml_parser_context_t* context, con
...
+            goto str_to_scalar_error;
```
filip-stenstrom, line 703: Keep `err2` and write the log the error here instead

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -664,51 +668,63 @@ static int fmi3_xml_str_to_array_floatXX(fmi3_xml_parser_context_t* context, con
...
-            if (fmi3_xml_str_to_floatXX(context, 1, writeAddr, NULL, token, primType)) {
-                jm_log_error(context->callbacks, module, "Unable to parse to %s: %s", primType->name, token);
-                res = -1;
-                goto err2;
+            if ((fmi3_base_type_enu_is_int(primType->baseType) &&
+                    fmi3_xml_str_to_intXX(context, 1, writeAddr, NULL, token, primType))
+                    ||
+                (fmi3_base_type_enu_is_float(primType->baseType) &&
+                    fmi3_xml_str_to_floatXX(context, 1, writeAddr, NULL, token, primType))) {
+                goto str_to_array_error;
```
filip-stenstrom, line 721: Same

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -743,13 +765,23 @@ int fmi3_xml_set_attr_array(fmi3_xml_parser_context_t *context, fmi3_xml_elm_enu
...
+            type_as_str = "float";
+        } else if(fmi3_base_type_enu_is_int(primType->baseType)) {
+            type_as_str = "int";
+        }
+        fmi3_xml_parse_error(context, "XML element '%s': could not parse value for %s attribute '%s'='%s'",
+            elmName,
+            type_as_str,
+            attrName,
+            str
+        );
```
filip-stenstrom, line 782: Here we have two args on same line, then one for each row? And sometimes ); on new line and sometimes not. I think we need to discuss how to format function calls.

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -743,13 +765,23 @@ int fmi3_xml_set_attr_array(fmi3_xml_parser_context_t *context, fmi3_xml_elm_enu
     }
 
     /* write all attributes to array of correct type */
-    if (fmi3_xml_str_to_array_floatXX(context, str, arrPtr, arrSize, primType)) {
+    if (fmi3_xml_str_to_array(context, str, arrPtr, arrSize, primType)) {
         jm_string elmName = fmi3_element_handle_map[elmID].elementName;
         jm_string attrName = fmi3_xmlAttrNames[attrID];
-        fmi3_xml_parse_error(context, "XML element '%s': could not parse value for float attribute '%s'='%s'", elmName, attrName, str);
+        char* type_as_str = "";
```
filip-stenstrom, line 771: Use `primType->name`?

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1168,36 +1240,49 @@ int fmi3_xml_handle_IntXXVariable(fmi3_xml_parser_context_t* context, const char
             type = (fmi3_xml_int_type_props_t*)declaredType;
         }
         variable->type = &type->super;
+        /* TODO handle this buffer clean-up in a better way */
+        {
+            const char* tmp; /* unused */
+            fmi3_xml_get_attr_str(context, fmi3_xml_elmID_Int64, fmi_attr_id_start, 0, &tmp);
+        }
```
filip-stenstrom, line 1247: We should handle this on all places before the ticket is done. We can save that for last but just keeping this thread open as reminder

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1229,40 +1314,55 @@ int fmi3_xml_handle_BooleanVariable(fmi3_xml_parser_context_t *context, const ch
...
+    } else {
+        /* We must wait until after parsing dimensions, because we can't otherwise know
+         * if it's a 1x1 array or a scalar variable just by reading the start value. */
+        hasStart = variable->startAttr != NULL;
         if(hasStart) {
-            fmi3_xml_variable_start_int_t * start = (fmi3_xml_variable_start_int_t*)fmi3_xml_alloc_variable_type_start(td, variable->type, sizeof(fmi3_xml_variable_start_int_t));
-            if(!start) {
+            jm_string startAttr = variable->startAttr;
+            fmi3_xml_variable_start_int_t* start = (fmi3_xml_variable_start_int_t*)fmi3_xml_alloc_variable_type_start(
+                    td, variable->type, sizeof(fmi3_xml_variable_start_int_t));
```
filip-stenstrom, line 1342: I think it would make sense to create fmi3_xml_variable_start_boolean_t by now

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -443,22 +443,24 @@ static int fmi3_xml_value_boundary_check_strcmp(jm_string strVal, const char* fo
...
 */
-static int fmi3_xml_str_to_bool(fmi3_xml_parser_context_t *context, int required, void* field, void* defaultVal,
-        jm_string strVal, const fmi3_xml_primitive_type_t* primType) {
-    /* Should we do something with the values 'required' and 'defaultVal' or can we simply remove them? */
+static int fmi3_xml_str_to_bool(fmi3_xml_parser_context_t *context, int required, bool* field, bool* defaultVal,
+        jm_string strVal) {
     int status = -1;
+
+    if (!strVal && !required) {
+        field = defaultVal;
```
filip-stenstrom, line 455: *field

and shouldn't there be a case for !strVal && required? The strcmp below will fail, so will fail, but explicit check kind of makes sense. You can probably use same code as I added in #39 (just search for 'required').

filip-stenstrom: Or rather:
*field = *defaultVal

Or why does defaultVal have to be a pointer here?

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -832,7 +832,9 @@ int fmi3_xml_handle_Variable(fmi3_xml_parser_context_t *context, const char* dat
         /* Save start value for processing after reading all Dimensions */
         /* variable->startAttr = fmi3_xml_peek_attr_str(context, fmi_attr_id_start);*/
         variable->startAttr = fmi3_xml_peek_attr_str(context, fmi_attr_id_start);
-
+        if (variable->vr == 118) {
+            printf("I am here\n");
+        }
```
filip-stenstrom, line 837: But not for long :)

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -673,42 +714,64 @@ static int fmi3_xml_str_to_array_floatXX(fmi3_xml_parser_context_t* context, con
...
     vals = context->callbacks->malloc(nVals * primType->size); /* freed in fmi3_xml_clear_model_description */
     if (!vals) {
         JM_LOG_ERROR_NO_MEM();
         res = -1;
         goto err2;
     }
 
-    /* write start value(s) to correct type */
+    /* Write start value(s) to correct type */
     if (nVals == 1) { /* scalar */
```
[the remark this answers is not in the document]
modelonrobinandersson: I believe this happens if Dimension is set to 1. We should test this however but I prefer to do it after implementing all the types.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -650,21 +716,28 @@ fmi3_xml_bool_variable_t* fmi3_xml_get_variable_as_boolean(fmi3_xml_variable_t*
...
-    case fmi3_base_type_int64:    /* fallthrough */
-    case fmi3_base_type_int32:    /* fallthrough */
-    case fmi3_base_type_int16:    /* fallthrough */
-    case fmi3_base_type_int8:     /* fallthrough */
-    case fmi3_base_type_uint64:   /* fallthrough */
-    case fmi3_base_type_uint32:   /* fallthrough */
-    case fmi3_base_type_uint16:   /* fallthrough */
-    case fmi3_base_type_uint8:    /* fallthrough */
+    case fmi3_base_type_uint64:
+        return fmi3_xml_get_int_variable_start((fmi3_xml_int_variable_t*)v).array64u;
```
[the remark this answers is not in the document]
modelonrobinandersson: I will also fix this after implementing all the types, I have noticed the unnecessary discrepancy too.

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -238,6 +238,13 @@ const fmi3_xml_primitive_types_t PRIMITIVE_TYPES = {
         0, /* N/A */
         0,
         fmi3_base_type_bool,
+    },
+    {
+        "Enumeration",
+        sizeof(fmi3_int32_t),
+        fmi3_bitness_32, /* TODO: Perhaps this should be 64bit in FMI3??? */
```
modelonrobinandersson, line 245: Note that the FMI standard says:
`To set or inquire variables of type [Enumeration](https://fmi-standard.org/docs/3.0/#figure-schema-Enumeration), fmi3SetInt64 and fmi3GetInt64 must be used.`

src/Util/src/FMI3/fmi3_enums.c
```diff
@@ -51,6 +51,14 @@ int fmi3_base_type_enu_is_float(fmi3_base_type_enu_t baseType) {
     return 0;
 }
 
+int fmi3_base_type_enu_is_enum(fmi3_base_type_enu_t baseType) {
+    switch (baseType) {
+        case fmi3_base_type_enum: return 1;
+        default: return 0;
+    }
+    return 0;
```
filip-stenstrom, line 59: Makes more sense to have an `if` here, or even better change it to return `bool` perhaps? Then `return baseType == fmi3_base_type_enum`

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -704,13 +768,31 @@ fmi3_xml_string_variable_t* fmi3_xml_get_variable_as_string(fmi3_xml_variable_t*
...
+fmi3_string_t* fmi3_xml_get_string_variable_start_array(fmi3_xml_string_variable_t* v){
+    fmi3_xml_variable_t* vv = (fmi3_xml_variable_t*)v;
+    if(fmi3_xml_get_variable_has_start(vv)) {
+        fmi3_xml_string_variable_start_t* start = (fmi3_xml_string_variable_start_t*)(vv->type);
+        return (fmi3_string_t*)jm_vector_get_itemp(jm_voidp)(&start->stringStartValues, 0);
+    }
+    return NULL;
+}
+
+size_t* fmi3_xml_get_string_variable_start_array_length(fmi3_xml_string_variable_t* v){
```
modelonrobinandersson, line 789: Remove?

src/XML/src/FMI3/fmi3_xml_parser.c
modelonrobinandersson: Fixed in #6ba08da

src/XML/src/FMI3/fmi3_xml_variable.c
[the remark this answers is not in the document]
modelonrobinandersson: Fixed

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -968,19 +965,85 @@ static void* fmi3_xml_get_variable_start_array(fmi3_xml_variable_t* v) {
...
         var->dimensionsArray = NULL;
+    } else {
+        /* Scalar String variables can have a start value that we need to free.*/
+        if (var->type->baseType == fmi3_base_type_str) {
+            fmi3_xml_string_variable_start_t* start = (fmi3_xml_string_variable_start_t*)(var->type);
+            if (&start->stringStartValues) {
+                jm_vector_foreach(jm_voidp)(&start->stringStartValues, callbacks->free);
+                //jm_vector_free_data(jm_voidp)(&start->stringStartValues);
+            }
+        }
```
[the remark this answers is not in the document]
modelonrobinandersson: We treat it differently compared to binary because Binary variables that are not arrays are using the old implementation (because of the reasons we discussed when we realized that errors during parsing are treated very differently from case to case).

jm_vector_free_data here causes `Invalid read of size 8` and since memcheck does not complain **without** `jm_vector_free_data` I was initially thinking the current implementation is correct.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -968,19 +965,85 @@ static void* fmi3_xml_get_variable_start_array(fmi3_xml_variable_t* v) {
...
+    return 0;
+}
+
 void fmi3_xml_variable_free_internals(jm_callbacks* callbacks, fmi3_xml_variable_t* var) {
+    /* We need special handling of Strings and Binaries because
+    *  of how start values are handled differently for those types.
+    */
     if (fmi3_xml_variable_is_array(var)) {
         if (var->type->baseType == fmi3_base_type_str) {
             fmi3_xml_string_variable_start_t* start = (fmi3_xml_string_variable_start_t*)(var->type);
```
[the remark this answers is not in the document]
modelonrobinandersson: Do you think it makes sense to add a check if `var` has a start value before we do the calls above?

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1842,30 +1950,32 @@ int fmi3_xml_handle_StringVariable(fmi3_xml_parser_context_t *context, const cha
...
-                jm_vector_init(jm_voidp)(&start_obj->arrayStartValues, 0, context->callbacks);
-                size_t n = jm_vector_copy(jm_voidp)(&start_obj->arrayStartValues, &context->variableStartValues);
-                variable->type = &start_obj->super;
-                size_t s = jm_vector_get_size(jm_voidp)(&context->variableStartValues);
-                if (s != n) {
+            } else {
+                if (nStart != 1) {
                     fmi3_xml_parse_fatal(context, "Could not retrieve string start values");
                     return -1;
                 }
```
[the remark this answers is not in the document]
modelonrobinandersson: What is the proposed change here? To me it seems reasonable to check that it `if (nStart != 1) ` because we know it is not an array. If `nStart != 1` evaluates to `true` it seems reasonable to break immediately.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1909,86 +2017,63 @@ int fmi3_xml_handle_Start(fmi3_xml_parser_context_t* context, const char* data)
...
+            if (len % 2 != 0) {
+                // 2 hexadecimal chars per byte. This is also checked in the conversion, but if this happens
+                // and we want to handle it, then we must allocate extra size to take into account padding of
+                // the final char.
+                return -1;
+            }
+            char* attrAsStr = context->callbacks->malloc(len + 1);
+            strcpy(attrAsStr, attr);
+            jm_vector_push_back(jm_voidp)(vec, attrAsStr);
+        } else {
```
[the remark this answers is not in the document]
modelonrobinandersson: Will fix this in a separate MR.

PR #44, opened 2023-05-04
Update valueRefs to be globally unique for FMI3

filip-stenstrom:


.vscode/launch.json
```diff
@@ -7,7 +7,7 @@
             "type": "cppdbg",
             "request": "launch",
             "program": "${command:cmake.launchTargetPath}",
-            "args": [],
+            // "args": ["/mnt/fmi-library/build_cmake/Testing/BouncingBall3_me"], <-- fmi3_import_xml_test_me
```
filip-stenstrom, line 10: Some tests are parametrized on the same .c file, so it's not trivial to fix. I think it's better to just add the args here until we can fix it.

Config.cmake/config_test.h.cmake
```diff
@@ -25,10 +25,8 @@
 /* PATHs to temporary test directory */
 #define FMU_TEMPORARY_TEST_DIR "@FMU_TEMPORARY_TEST_DIR@"
 
-#define COMPRESS_DUMMY_FILE_PATH_SRC "@COMPRESS_DUMMY_FILE_PATH_SRC@"
-#define COMPRESS_DUMMY_FILE_PATH_DIST "@COMPRESS_DUMMY_FILE_PATH_DIST@"
-#define UNCOMPRESSED_DUMMY_FILE_PATH_SRC "@UNCOMPRESSED_DUMMY_FILE_PATH_SRC@"
-#define UNCOMPRESSED_DUMMY_FOLDER_PATH_DIST "@UNCOMPRESSED_DUMMY_FOLDER_PATH_DIST@"
+#define TEST_OUTPUT_FOLDER "@TEST_OUTPUT_FOLDER@"
+#define FMIL_TEST_DIR "@FMIL_TEST_DIR@"
```
filip-stenstrom, line 29: Moving code from cmake to the .c/.cpp test files. It's easier to build the paths in C and doesn't clutter the cmake code far from where it's used.

Config.cmake/runtime_test.cmake
```diff
@@ -82,29 +83,22 @@ function(to_native_c_path path native_c_path)
...
-#Create paths for the config_test.h
-if(WIN32)
-    set(DLL_OUTPUT_PATH "${TEST_OUTPUT_FOLDER}")
-
-    string(REPLACE "/" "\\\\" UNCOMPRESSED_DUMMY_FILE_PATH_SRC "${UNCOMPRESSED_DUMMY_FILE_PATH_SRC}")
-    string(REPLACE "/" "\\\\" UNCOMPRESSED_DUMMY_FILE_PATH_DIST "${UNCOMPRESSED_DUMMY_FILE_PATH_DIST}")
-    string(REPLACE "/" "\\\\" UNCOMPRESSED_DUMMY_FOLDER_PATH_DIST "${UNCOMPRESSED_DUMMY_FOLDER_PATH_DIST}")
-    string(REPLACE "/" "\\\\" COMPRESS_DUMMY_FILE_PATH_SRC "${COMPRESS_DUMMY_FILE_PATH_SRC}")
-    string(REPLACE "/" "\\\\" COMPRESS_DUMMY_FILE_PATH_DIST "${COMPRESS_DUMMY_FILE_PATH_DIST}")
-endif(WIN32)
```
filip-stenstrom, line 103: Moved all of it to the .c files. Really want to keep cmake code short and readable.

Config.cmake/runtime_test.cmake
```diff
@@ -82,29 +83,22 @@ function(to_native_c_path path native_c_path)
...
+# (fmiXX_import_xml_test_XX) point at just for testing an XML. They should
+# be fixed to point at XML_PATH_T.
+#
+# @OUTPUT_FOLDER_T:        The directory which the FMU will be placed in. (The uncompressed directory is also created here.)
+# @MODEL_IDENTIFIER_T:     The model identifier of the FMU. Will be used as stem for the FMU name.
+# @FILE_NAME_CS_ME_EXT_T:  A suffix added to the stem of the FMU base name.
+# @TARGET_NAME_T:          The target for building the shared library.
+# @XML_PATH_T:             The path to the modelDescription that will be zipped.
+# @SHARED_LIBRARY_PATH_T:  The path to the shared library produced by TARGET_NAME_T.
 function(compress_fmu OUTPUT_FOLDER_T MODEL_IDENTIFIER_T FILE_NAME_CS_ME_EXT_T TARGET_NAME_T XML_PATH_T SHARED_LIBRARY_PATH_T)
```
filip-stenstrom, line 99: Only slightly simplified and documented this function. No changes.

Test/FMI3/fmi3_capi_basic_test.cpp
```diff
@@ -21,18 +21,6 @@
...
-} fmi3_inst_env_t;
-
-
-/* The following dummy types/functions are used only for testing in order to be able
-    invoke the necessary functions, for example FMU instantiation.
-*/
-void fmi3_dummy_log_callback(fmi3_instance_environment_t instEnv, fmi3_status_t status,
-        fmi3_string_t category, fmi3_string_t message) {
-    fmi3_log_forwarding(((fmi3_inst_env_t*)instEnv)->fmu, status, category, message);
-}
```
filip-stenstrom, line 35: Unnecessary complexity, and is not what we test, so just removed. The log callback is tested elsewhere.

src/Import/src/FMI3/fmi3_import_capi.c
```diff
@@ -195,6 +195,13 @@ jm_status_enu_t fmi3_import_instantiate_model_exchange(fmi3_import_t* fmu,
     fmi3_instance_t c;
     if (!resourcePath)
         resourcePath = fmu->resourcePath;
+    
+    fmu->instanceName = fmu->callbacks->malloc(strlen(instanceName)+1);
+    if (!fmu->instanceName) {
+        jm_log_fatal(fmu->callbacks, module, "Could not allocate memory");
+        return fmi3_status_fatal;
+    }
+    strcpy(fmu->instanceName, instanceName);
```
filip-stenstrom, line 204: Fixes the TODO in fmi3_log_forwarding (the problem is that FMI2 logger callback has `const char* instanceName` arg, but that's gone in FMI3, so we need to save it at instantiation instead to keep old behavior).

src/Import/src/FMI3/fmi3_import_convenience.c
```diff
@@ -274,17 +252,21 @@ void fmi3_log_forwarding(fmi3_instance_environment_t inst, fmi3_status_t status,
...
-         buf = jm_vector_get_itemp(char)(&fmu->logMessageBufferCoded,0);
-    }
-    else  {
+    if (fmu) {
+        cb = fmu->callbacks;
+        buf = jm_vector_get_itemp(char)(&fmu->logMessageBufferCoded, 0);
+    } else {
+        // XXX: Why do even accept that fmu==NULL? What's the point of forwarding
+        // the logging to the FMU in that case? Also, it's not documented in the
+        // function's interface.
```
filip-stenstrom, line 263: I rewrote the function's documentation now... But the old one didn't mention it either, and it really doesn't make sense. We can probably remove it, but preferred to not mess with it right now.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -2079,6 +2079,16 @@ int fmi3_xml_handle_ModelVariables(fmi3_xml_parser_context_t *context, const cha
...
             }
             jm_vector_qsort(jm_voidp)(md->variablesByVR, fmi3_xml_compare_vr_and_original_index);
+
+            for (size_t i = 0; i < size-1; ++i) {
+                fmi3_xml_variable_t* v1 = jm_vector_get_item(jm_voidp)(md->variablesByVR, i);
+                fmi3_xml_variable_t* v2 = jm_vector_get_item(jm_voidp)(md->variablesByVR, i+1);
+                if (v1->vr == v2->vr) {
+                    fmi3_xml_parse_error(context, "The following variables have the same value reference: %s, %s",
+                            v1->name, v2->name);
+                    // Usually we return -1 on error, but should be fine to continue.
```
filip-stenstrom, line 2089: @PeterMeisrimelModelon
It's not really clear how we should handle parse errors (in general). In this case I thought it overly dramatic to stop the whole ModelVariable handler, because that will leave everything broken. I know other code in same function do it, but unclear why.

PeterMeisrimelModelon (commented):
Looks good to me, a few comments, which are mostly for my understanding.

For the typo fix in the linked issue. Shouldn't we add some sort of log message or printout about the deprecation there, such we can actually remove the deprecated one at some point in the future?

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: Partially. The problem is if we give a fatal error then we don't return anything from the parse call. I.e. returned fmi3_import_t* will be NULL. So let's say you just wanted to display the parsed variables in a GUI, then you can't even do that.
However, it may lead to secondary errors when parsing ModelStructure, or ambiguities when trying to retrieve the variable with that VR.

So it's not clear what is best, and we don't have any clear guidelines of when something should be fatal. But I could go with parse_fatal in this case. It's easier to relax that restriction than going the other way.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1254,7 +1289,7 @@ int fmi3_xml_handle_Dimension(fmi3_xml_parser_context_t* context, const char* da
 
 int fmi3_xml_handle_FloatXXVariable(fmi3_xml_parser_context_t* context, const char* data,
         fmi3_xml_float_type_props_t* defaultType,
-        fmi3_xml_elm_enu_t elmID, /* ID of the Type (not the Variable) */
+        fmi3_xml_elm_enu_t elmID, /* ID of the Type (not the Variable) */  // XXX: Why?
```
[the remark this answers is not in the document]
filip-stenstrom: From what I could see briefly was that it's only used for error reporting. The point is that it's confusing to give the element ID of the type when you're inside the variable. So if possible, we should fix that. Or maybe it's not even needed anymore. At least it should be motivated why. I also didn't want to start digging in floatXX code as part of this ticket especially when RA is working on it.

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: So my proposal is to change to parse_fatal, would that be OK?

src/XML/src/FMI3/fmi3_xml_variable.c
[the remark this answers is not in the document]
PeterMeisrimelModelon: Agree, this could and should probably be improved, but at some later point.

src/XML/src/FMI3/fmi3_xml_variable.c
PeterMeisrimelModelon: If I understand this correctly, fmi3_xml_parse_fatal would stop parsing, give and appropriate log message, but you would still get the parsed result until that point?

filip-stenstrom: You only get the log.

`fmi3_xml_parse_fatal` calls `XML_StopParser(context->parser,0);`, where arg2==0 which means "won't resume".

`fmi3_import_parse_xml` calls `fmi3_xml_parse_model_description` calls `XML_Parse`. All our handlers are invoked from that function. Inside that function we call parse_fatal, which leads to `XML_Parse` returning some ERROR enum, because we called StopParser with "won't resume", which in turn leads to us eventually returning NULL for the whole parse call.

PeterMeisrimelModelon: Ah, thanks for clarifying. Sounds good to me.

PeterMeisrimelModelon (approved):

PR #46, opened 2023-05-09
Parse TypeDefinition for Binary and Clock

filip-stenstrom:


src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -380,7 +380,7 @@ int fmi3_xml_set_attr_string(fmi3_xml_parser_context_t *context, fmi3_xml_elm_en
 int fmi3_xml_set_attr_enum(fmi3_xml_parser_context_t *context, fmi3_xml_elm_enu_t elmID, fmi3_xml_attr_enu_t attrID,
         int required, unsigned int* field, unsigned int defaultVal, jm_name_ID_map_t* nameMap)
 {
-    int ret, i;
+    int i = 0;
```
filip-stenstrom, line 383: I'll get a conflict I know...

src/XML/src/FMI3/fmi3_xml_type.c
```diff
@@ -68,245 +68,250 @@ fmi3_xml_int_typedef_t* fmi3_xml_get_type_as_int(fmi3_xml_variable_typedef_t* t)
...
+// -----------------------------------------------------------------------------
+// Float64
+// -----------------------------------------------------------------------------
+
+static fmi3_xml_float_type_props_t* fmi3_xml_get_float_type_props(fmi3_xml_float_typedef_t* t) {
     fmi3_xml_variable_typedef_t* vt = (void*)t;
-    fmi3_xml_float_type_props_t* props = (fmi3_xml_float_type_props_t*)(vt->super.nextLayer);
-    return props->typeMin.scalar64s;
+    return (void*)(vt->super.nextLayer);
+}
```
filip-stenstrom, line 132: Helper function just to reduce code duplication

src/XML/src/FMI3/fmi3_xml_type.c
```diff
@@ -637,15 +700,21 @@ int fmi3_xml_handle_SimpleType(fmi3_xml_parser_context_t *context, const char* d
...
-        fmi3_xml_variable_type_base_t* base, size_t typeSize)
+        fmi3_xml_variable_type_base_t* nextLayer, size_t typeSize)
 {
     jm_callbacks* cb = td->typeDefinitions.callbacks;
     fmi3_xml_variable_type_base_t* type = cb->malloc(typeSize);
-    if (!type) return NULL;
-    fmi3_xml_init_variable_type_base(type, fmi3_xml_type_struct_enu_props, base->baseType);
-    type->nextLayer = base;
+    if (!type) {
+        jm_log_fatal(cb, module, "Could not allocate memory"); // TODO: Change to fmi3_xml_parse_fatal
```
filip-stenstrom, line 708: Would cause quite some conflicts with array ticket if I did this TODO now, because it requires 'context'.

(I also want to rename this function. It does a lot more than just allocates.)

src/XML/src/FMI3/fmi3_xml_type.c
```diff
@@ -858,74 +951,217 @@ int fmi3_xml_handle_UInt8(fmi3_xml_parser_context_t* context, const char* data)
...
-        type->super.nextLayer = &md->typeDefinitions.defaultStringType;
-    }
-    else {
-        /* don't do anything. might give out a warning if(data[0] != 0) */
-        return 0;
+        // Get the typedef already created for SimpleType and set props
+        typeDef = jm_vector_get_last(jm_named_ptr)(&context->modelDescription->typeDefinitions.typeDefinitions).ptr;
+        typeDef->super.baseType = fmi3_base_type_str;
+        typeDef->super.nextLayer = &md->typeDefinitions.defaultStringType;
     }
```
filip-stenstrom, line 1154: Diff just looks strange. I changed _handle_SimpleType to be a one-liner and renamed type -> typeDef.

src/XML/src/FMI3/fmi3_xml_type_impl.h
```diff
@@ -116,16 +132,19 @@ typedef enum {
...
-
-    fmi3_xml_variable_type_base_t* next;        /* should only be used for deallocation */
+    fmi3_xml_variable_type_base_t* nextLayer;   /* The next layer in the type aggregate */ // TODO: Rename to next
+    fmi3_xml_type_struct_kind_enu_t structKind; /* The actual (sub) type */
+    fmi3_base_type_enu_t baseType;              /* The FMI base type */
+    char isRelativeQuantity;                    /* RelativeQuantity flag (only used in fmi3_xml_real_type_props_t) */
+    char isUnbounded;                           /* Unbounded flag        (only used in fmi3_xml_real_type_props_t) */
+
+    // TODO: Remove (create a new jm_vector on typeDefinitions that works the same)
+    fmi3_xml_variable_type_base_t* next;        /* For deallocation: the next node in the deallocation list */
```
filip-stenstrom, line 147: Would also cause array ticket conflicts. And make this ticket harder to review than it already is.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1322,46 +1322,24 @@ int fmi3_xml_handle_FloatXXVariable(fmi3_xml_parser_context_t* context, const ch
...
-                }
-
-                type->super.nextLayer = declaredType;
-                if (!hasUnit) type->displayUnit              = dtProps->displayUnit;
-                if (!hasMin)  type->typeMin                  = dtProps->typeMin;
-                if (!hasMax)  type->typeMax                  = dtProps->typeMax;
-                if (!hasNom)  type->typeNominal              = dtProps->typeNominal;
-                if (!hasQuan) type->quantity                 = dtProps->quantity;
-                if (!hasRelQ) type->super.isRelativeQuantity = dtProps->super.isRelativeQuantity;
-                if (!hasUnb)  type->super.isUnbounded        = dtProps->super.isUnbounded;
```
filip-stenstrom, line 1361: All of this is handled in the the parse_float_type_properties. Before it used the default type for non-set optional attrs. Now it immediately uses the declaredType.

Test/FMI3/fmi3_import_type_definitions_test.cpp
```diff
@@ -0,0 +1,1393 @@
...
+#define UINTXX_MIN (0)
+
+/* delta offsets from extreme values */
+#define D_TYPE  (1)  /* for types when overriding defaults */
+#define D_VAR   (2)  /* for variables when overriding defaults/typedefs */
+#define D_START (3)  /* for variable start values when overriding defaults */
+
+static int G_nErrors; /* global used to count the number of errors reported to the logger during parsing - reset before use! */
+
+static fmi3_import_variable_typedef_t* get_typedef_by_name(fmi3_import_type_definitions_t* tds, const char* name) {
```
PeterMeisrimelModelon, line 20: Not sure how useful this would be in practice, but this almost feels like something that could also be part of the API.

Test/FMI3/fmi3_import_type_definitions_test.cpp
```diff
@@ -0,0 +1,1393 @@
...
+        REQUIRE(fmi3_import_get_variable_declared_type(v) == nullptr);
+        REQUIRE(strcmp(mimeType, mimeTypeDefault) == 0);
+        REQUIRE(maxSize == maxSizeDefault);
+    }
+    SECTION("Variable inheriting minimal typedef") {
+        v = fmi3_import_get_variable_by_name(xml, "var_inh_minimal_1");
+        bv = fmi3_import_get_variable_as_binary(v);
+        mimeType = fmi3_import_get_binary_variable_mime_type(bv);
+        maxSize  = fmi3_import_get_binary_variable_max_size(bv);
+        REQUIRE(fmi3_import_get_variable_declared_type(v) == (void*)tdMinimal);
```
PeterMeisrimelModelon, line 1173: What is the purpose of the (void*) here? This appears in a bunch of places and I don't quite understand it, since the return type of fmi3_import_get_variable_declared_type matches up with the type of here tdMinimal.

PeterMeisrimelModelon (commented):
Looks great, no complaints, mostly a few questions, see comments.

Test/FMI3/fmi3_import_type_definitions_test.cpp
filip-stenstrom: No, they are different. LHS is `fmi3_import_variable_typedef_t*` and RHS is `fmi3_import_binary_typedef_t*`. In this specific test I could've used the API call that converts to the specific type's typedef, but I kind of took a shortcut since all that functions just do a typecast.

You can assign or compare a void* with any kind of void pointer, so that's a short way of not needing to write the whole super long type name everywhere.

On the places where it's assignment, casting to void* is the same as casting to whatever the type is of the declared variable. Just less text, and no need to change the typecast if the declared type changes.

Test/FMI3/parser_test_xmls/type_definitions/valid/clock/modelDescription.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<fmiModelDescription fmiVersion="3.0" modelName="" instantiationToken="">
+
+<ModelExchange modelIdentifier="myModelIdentifier" />
+
+<TypeDefinitions>
+  <!-- NOTE:
+    It doesn't make much sense to put intervalVariability on typedefs when they must be
```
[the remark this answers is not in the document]
filip-stenstrom: Yes that was what I meant.

And yes we probably should do that. I'll look through spec and issues one more time first though.

src/XML/src/FMI3/fmi3_xml_type.c
filip-stenstrom: See next

src/XML/src/FMI3/fmi3_xml_type.c
```diff
@@ -68,245 +68,250 @@ fmi3_xml_int_typedef_t* fmi3_xml_get_type_as_int(fmi3_xml_variable_typedef_t* t)
...
+    }
     return du;
 }
 
+// -----------------------------------------------------------------------------
+// Int64
+// -----------------------------------------------------------------------------
 
-fmi3_int64_t fmi3_xml_get_int64_type_min(fmi3_xml_int_typedef_t* t) {
+static fmi3_xml_int_type_props_t* fmi3_xml_get_int_type_props(fmi3_xml_int_typedef_t* t) {
```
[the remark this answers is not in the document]
filip-stenstrom: Otherwise I'll need to move it way up above FloatXX, which would break the pattern of having it under the "header" like for the others. I could also make an IntXX one, but the headers are mainly to allow some quick visual separation of the code, given its repetitive nature.

src/XML/src/FMI3/fmi3_xml_type.c
```diff
@@ -768,59 +852,68 @@ int fmi3_xml_handle_Float32(fmi3_xml_parser_context_t* context, const char* data
...
     fmi3_xml_model_description_t* md = context->modelDescription;
     fmi3_xml_type_definitions_t* td = &md->typeDefinitions;
-    fmi3_xml_int_type_props_t* props = 0;
+    fmi3_xml_int_type_props_t* fallbackProps = NULL;
+    fmi3_xml_int_type_props_t* props = NULL;
     const char* quantity = NULL;
 
-    jm_vector(char)* bufQuantity = fmi3_xml_reserve_parse_buffer(context, 3, 100);
-    if (!bufQuantity) return 0;
+    // XXX: Why don't we give fatal errors in this funciton, but we do it for floatXX?
```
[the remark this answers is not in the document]
filip-stenstrom: We have a discussion about this on Wed. Let's save it for then

Test/FMI3/fmi3_import_type_definitions_test.cpp
filip-stenstrom: I don't know either. But I don't want to push for new API unless requested or if it would be a large performance gain

filip-stenstrom:
> Looks great, no complaints, mostly a few questions, see comments.
> 
> Note: For the tests, I only checked the binary and clocked tests and take it that the remaining ones are a straight-forward but tedious conversion to catch2. Please notify if otherwise.

Correct

Test/FMI3/fmi3_import_type_definitions_test.cpp
PeterMeisrimelModelon: Ok.

Test/FMI3/fmi3_import_type_definitions_test.cpp
PeterMeisrimelModelon: Ok, thanks for clarifying.

Test/FMI3/parser_test_xmls/type_definitions/valid/clock/modelDescription.xml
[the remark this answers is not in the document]
filip-stenstrom: As said on daily, let's hold off on this for a while

src/XML/src/FMI3/fmi3_xml_type.c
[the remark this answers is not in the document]
filip-stenstrom: @PeterMeisrimelModelon So this should be the only thing left. I'd prefer to not delay merging for this. What do you think about me having an action to fix at least this TODO after the meeting, and get this merged now?

PR #57, opened 2023-05-31
Parse Alias elements

filip-stenstrom:
Alias are now specified by the Alias element. This PR parses it and adds new API.

* Aliases now do not result in `fmi3_import_variable_t`, but their data as specified in XML is instead retrievable from the base variable
* Alias base variables can also be retrieved from `fmi3_import_get_variable_by_name` by passing alias name


Test/FMI3/fmi3_import_convenience_test.cpp
```diff
@@ -168,18 +168,6 @@ TEST_CASE("Variable name expansion via logger") {
     fmi3_testutil_import_free(tfmu);
 }
 
-TEST_CASE("Error check: Variables with same VR") {
```
filip-stenstrom, line 171: Removed since we have equivalent and more in variables_test.cpp

Test/FMI3/fmi3_import_variable_test.cpp
```diff
@@ -118,7 +118,7 @@ static void test_binary_all_attrs(fmi3_import_t* xml) {
     REQUIRE(fmi3_import_get_variable_vr(v) == 4);
     REQUIRE(fmi3_import_get_causality(v) == fmi3_causality_enu_output);
     REQUIRE(fmi3_import_get_variability(v) == fmi3_variability_enu_discrete);
-    REQUIRE(fmi3_import_get_initial(v) == fmi3_initial_enu_exact);
+    REQUIRE(fmi3_import_get_initial(v) == fmi3_initial_enu_calculated);
```
filip-stenstrom, line 121: Was giving error because exact requires a start value, so I just changed to calculated

Test/FMI3/parser_test_xmls/convenience/valid/get_variable_by_alias_name1/modelDescription.xml
```diff
@@ -1,12 +1,12 @@
...
 <ModelExchange modelIdentifier="myModelIdentifier" />
 
 <ModelVariables>
-  <!-- Variables with same VR -->
-  <Int32 name="v1" valueReference="1"/>
-  <Int32 name="v2" valueReference="2"/>
-  <Int32 name="v3" valueReference="1"/>
+    <Float64 name="v1" valueReference="1">
+        <Alias name="v1_a1"/>
+    </Float64>
```
filip-stenstrom, line 9: Git believes this file was moved (the one from convenience). Ignore the diff, it's all new.

src/Util/include/JM/jm_named_ptr.h
```diff
@@ -59,7 +59,24 @@ The "name" is copied into the allocated memory.
...
+ * The allocated memory must be able to store 'name', and 'nameoffset' must give
+ * the offset to that address.
+ * 
+ * @param name:
+ *      The name.
+ * @param size:
+ *      Size of memory for the .ptr (just like for malloc). Do not add the size for the name.
+ * @param nameoffset:
+ *      Write the name at offset from the start of the requested 'size'.
+ */
```
filip-stenstrom, line 79: Mainly moved from .c to .h file to make intellisense show it

src/Util/include/JM/jm_vector.h
```diff
@@ -145,20 +145,25 @@ Default definition below is jm_diff and is implemented as (int)(first-second)
...
+  \brief Uses linear search to find items in a vector. JM_COMPAR_OP is used for comparison.
+
+  @param a - the vector;
+  @param item - the searched item;
 
-  T* jm_vector_find(T)(jm_vector(T)* a, T item, jm_compare_ft f)
+  Return:
+    T* jm_vector_find(T)(jm_vector(T)* a, T item, jm_compare_ft f)  returns a pointer to the found item or NULL if not found
+*/
+#define jm_vector_find(T) jm_mangle(jm_vector_find, T)
```
filip-stenstrom, line 156: It's a bit hard to see from the diff, but it looked like this:

```
/**
 * Doc for macro1
 * Doc for macro2
 * More doc for macro1
 * More doc for macro2
 * /
#define macro1
#define macro2
```

So I split it

src/Util/src/JM/jm_named_ptr.c
```diff
@@ -38,26 +38,6 @@ jm_named_ptr jm_named_alloc(const char* name, size_t size, size_t nameoffset, jm
...
- *      The name.
- * @param size:
- *      Size of memory for the .ptr (just like for malloc). Do NOT take into account size
- *      for the name since this function handles that.
- * @param nameoffset:
- *      Write the name at offset from the start of the requested 'size'.
- *      Impl. note: this needs to be calculated by caller, because it's possible
- *      to get different byte padding of the struct, and it allows the function
- *      to be used without structs that end with a name field.
- */
```
filip-stenstrom, line 60: Moved as mentioned in prev comment

src/XML/src/FMI3/fmi3_xml_model_description.c
```diff
@@ -92,10 +92,8 @@ fmi3_xml_model_description_t * fmi3_xml_allocate_model_description( jm_callbacks
     fmi3_xml_init_type_definitions(&md->typeDefinitions, cb);
 
     jm_vector_init(jm_named_ptr)(&md->variablesByName, 0, cb);
-
-    md->variablesOrigOrder = 0;
-
-    md->variablesByVR = 0;
+    jm_vector_init(jm_voidp)(&md->variablesOrigOrder, 0, cb);
+    jm_vector_init(jm_voidp)(&md->variablesByVR, 0, cb);
```
filip-stenstrom, line 96: Refactor: I saw no reason to have them being pointer, so made them non-pointer members

src/XML/src/FMI3/fmi3_xml_model_description.c
```diff
@@ -159,23 +157,15 @@ void fmi3_xml_clear_model_description( fmi3_xml_model_description_t* md) {
...
-        /* free extra pointers */
-        if (md->variablesOrigOrder) {
-            jm_vector_free(jm_voidp)(md->variablesOrigOrder);
-            md->variablesOrigOrder = 0;
-        }
-        /* free extra pointers */
-        if (md->variablesByVR) {
-            jm_vector_free(jm_voidp)(md->variablesByVR);
-            md->variablesByVR = 0;
+        for (i = 0; i < jm_vector_get_size(jm_voidp)(&md->variablesOrigOrder); i++) {
```
filip-stenstrom, line 160: variablesOrigOrder is now the list owning the memory, so had to swap things around a bit since it's not a jm_named_ptr but a jm_voidp

src/XML/src/FMI3/fmi3_xml_model_description.c
```diff
@@ -843,12 +833,8 @@ fmi3_xml_variable_t* fmi3_xml_get_variable_by_vr(fmi3_xml_model_description_t* m
     fmi3_xml_variable_t *pkey = &key;
     void** found;
 
-    if (!md->variablesByVR) return NULL;
-
-    // TODO: When variablesByVR no longer contains alias variables, then just return by index.
```
filip-stenstrom, line 848: This comment was false. You can't just do that (since VRs don't have to match the variable index)

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1083,14 +1063,21 @@ void fmi3_xml_variable_free_internals(jm_callbacks* callbacks, fmi3_xml_variable
...
+        size_t nAlias = jm_vector_get_size(jm_voidp)(var->aliases);
+        for (size_t i = 0; i < nAlias; i++) {
+            fmi3_xml_free_alias(callbacks, (fmi3_xml_alias_variable_t*)jm_vector_get_item(jm_voidp)(var->aliases, i));
+        }
+        jm_vector_free(jm_voidp)(var->aliases);
+        var->aliases = NULL;
     }
+
     if (var->clocks) {
-        callbacks->free(var->clocks);
```
filip-stenstrom, line 1091: Diff looks weird, but essentially we tried to free clocks 2 times (but set it to NULL after first, so never happened), so I removed one.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1306,63 +1315,60 @@ int fmi3_xml_handle_Variable(fmi3_xml_parser_context_t* context, const char* dat
...
-        /* Add Variable ptr to modelDescription obj */
-        named.ptr = 0;
-        named.name = 0;
-        pnamed = jm_vector_push_back(jm_named_ptr)(&md->variablesByName, named);
-        if (!pnamed) {
-            fmi3_xml_parse_fatal(context, "Could not allocate memory");
-            return -1;
+        // Create the alias and set name at same time:
+        const char* name;
+        if (jm_vector_get_size(char)(bufName) == 0) {  // FIXME: Empty string has been resized to 0
```
filip-stenstrom, line 1343: It's resized to 0 as part of `fmi3_xml_set_attr_string`, which is called from everywhere. I'm not fixing that now

src/XML/src/FMI3/fmi3_xml_variable_impl.h
```diff
@@ -36,6 +36,16 @@ typedef union fmi3_xml_valueref_or_variable_union_t {
     fmi3_xml_variable_t* variable;
 } fmi3_xml_valueref_or_variable_union_t;
 
+struct fmi3_xml_alias_variables_t {
+    jm_vector(jm_voidp) vec;
+};
```
filip-stenstrom, line 41: Needed to get opaque pointer in API working

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -1450,7 +1456,7 @@ int fmi3_xml_handle_Dimension(fmi3_xml_parser_context_t* context, const char* da
 
     if (!data) { /* start of tag */
 
-        fmi3_xml_variable_t* currentVar = jm_vector_get_last(jm_named_ptr)(&context->modelDescription->variablesByName).ptr;
+        fmi3_xml_variable_t* currentVar = jm_vector_get_last(jm_voidp)(&context->modelDescription->variablesOrigOrder);
```
filip-stenstrom, line 1459: Just a comment, but I'm quite happy with getting last variable from `variablesOrigOrder` instead. Makes a lot more sense

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -2160,169 +2205,80 @@ int fmi3_xml_handle_ModelVariables(fmi3_xml_parser_context_t *context, const cha
...
+            named.ptr = v;
+            jm_vector_push_back(jm_named_ptr)(&md->variablesByName, named);
+            if (v->aliases) {
+                fmi3_xml_alias_variable_t* alias;
+                for (size_t j = 0; j < jm_vector_get_size(jm_voidp)(v->aliases); j++) {
+                    alias = jm_vector_get_item(jm_voidp)(v->aliases, j);
+                    named.name = alias->name;
+                    named.ptr = v;
+                    jm_vector_push_back(jm_named_ptr)(&md->variablesByName, named);
+                }
```
filip-stenstrom, line 2228: (I could've just added the aliases to variablesByName after creating the other lists, but it made more sense to me clean it up while at it, i.e. that we add to variablesOrigOrder when we parse, then create the help lists.)

src/Import/include/FMI3/fmi3_import_variable.h
```diff
@@ -84,6 +91,17 @@ typedef struct fmi3_xml_clock_variable_t fmi3_import_clock_variable_t;
 typedef struct fmi3_import_variable_list_t fmi3_import_variable_list_t;
 /**@} */
 
+/**
+ * @name Types encapsulating alias information
+ * @{
+ */
+/** \brief List of the alias variables for a non-alias variable */
+typedef struct fmi3_xml_alias_variables_t fmi3_import_alias_variables_t;
```
PeterMeisrimelModelon, line 99: Maybe a more distinct name such as `fmi3_xml_alias_variables_list_t `?
Too easy to mix up with `fmi3_xml_alias_variable_t `

filip-stenstrom: Arrays, lists, and what not are troublesome when part of the API. While I would prefer `_list_t` I decided to not do it due to conventions.

```
FMILIB_EXPORT fmi3_import_unit_definitions_t* fmi3_import_get_unit_definitions(fmi3_import_t* fmu);
FMILIB_EXPORT fmi3_import_type_definitions_t* fmi3_import_get_type_definitions(fmi3_import_t* fmu);
```
And then we have more advanced lists for variables. See `fmi3_import_variable_list.h`. They have suffix `_list_t` so didn't want to mix them up.

But I would be happy with renaming to list if we could rename `fmi3_import_get_type_definitions` etc. (and also fix dimensions vector to be similar).

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -2160,169 +2205,80 @@ int fmi3_xml_handle_ModelVariables(fmi3_xml_parser_context_t *context, const cha
...
             }
         }
-
-        /* sort the variables by names */
         jm_vector_qsort(jm_named_ptr)(&md->variablesByName, jm_compare_named);
```
PeterMeisrimelModelon, line 2231: Not sure if we want to add this in this ticket, but once sorted, it would seem reasonably easy to add the global error check for unique names for variables & aliases.

src/Import/include/FMI3/fmi3_import.h
```diff
@@ -244,17 +244,6 @@ FMILIB_EXPORT fmi3_import_type_definitions_t* fmi3_import_get_type_definitions(f
 /** \brief Get a list of all the unit definitions in the model. */
 FMILIB_EXPORT fmi3_import_unit_definitions_t* fmi3_import_get_unit_definitions(fmi3_import_t* fmu);
 
-/** \brief Get the variable with the same value reference that is not an alias*/
-FMILIB_EXPORT fmi3_import_variable_t* fmi3_import_get_variable_alias_base(fmi3_import_t* fmu, fmi3_import_variable_t* v);
```
PeterMeisrimelModelon, line 248: Are we sure on removing this, I would think it could be worthwhile to have a such that function that returns the base variable, given an alias. Sure, one can do it via `fmi3_import_get_variable_by_name(fmi3_import_get_alias_variable_name(alias))`, but that feels stupidly expensive for large models.

Or is this functionality available somewhere else and I missed it?

src/Import/include/FMI3/fmi3_import_variable.h
PeterMeisrimelModelon: I'd be happy to rename all lists to `*_list_t` for consistency, probably in a follow up issue. 

I cannot really of another way to distinguish it by naming otherwise right now, without breaking the consistency with respect to `fmi3_xml_alias_variable_t`.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -949,6 +886,43 @@ fmi3_uint64_t fmi3_xml_get_clock_variable_shift_counter(fmi3_xml_clock_variable_
     return props->shiftCounter;
 }
 
+// -----------------------------------------------------------------------------
+// Alias variable
+// -----------------------------------------------------------------------------
+
+fmi3_xml_alias_variables_t* fmi3_xml_get_variable_aliases(fmi3_xml_variable_t* v) {
```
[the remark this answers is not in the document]
filip-stenstrom: We probably do. We'd need to go through all of API in that case. Adding OneNote bullet for it as well, since it's a design question.

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -949,6 +886,43 @@ fmi3_uint64_t fmi3_xml_get_clock_variable_shift_counter(fmi3_xml_clock_variable_
...
+
+size_t fmi3_xml_get_alias_variables_number(fmi3_xml_alias_variables_t* aliases) {
+    if (aliases) {
+        return jm_vector_get_size(jm_voidp)(&aliases->vec);
+    }
+    return 0;
+}
+
+fmi3_xml_alias_variable_t* fmi3_xml_get_alias(fmi3_xml_alias_variables_t* aliases, size_t index) {
+    if (!aliases || index > jm_vector_get_size(jm_voidp)(&aliases->vec)) {
```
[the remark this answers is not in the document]
filip-stenstrom: There isn't, but thinking again I don't think we should check for index here, but rather let it crash. That's the way that lists usually handle it (probably due to performance).

I mean, imagine we just returned an array instead, then it would segfault if it's NULL or if index is out of bounds, because it's a programming error

src/XML/src/FMI3/fmi3_xml_variable.c
```diff
@@ -2079,61 +2085,100 @@ int fmi3_xml_handle_EnumerationVariable(fmi3_xml_parser_context_t *context, cons
...
+        fmi3_xml_model_description_t* md = context->modelDescription;
+        fmi3_xml_variable_t* baseVar = jm_vector_get_last(jm_voidp)(&md->variablesOrigOrder);
+        size_t bufIdx = 1;
+
+        // FIXME: bufDesc will have size==0 when the attribute value is the empty string,
+        // making it indistinguishable from not being defined at all.
+        bool hasDesc = fmi3_xml_peek_attr_str(context, fmi_attr_id_description) != NULL;
+        
+        // Read the attributes to memory owned by FMIL:
+        jm_vector(char)* bufName = fmi3_xml_reserve_parse_buffer(context, bufIdx++, 100);
```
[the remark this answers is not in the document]
filip-stenstrom: It's the initial size. `fmi3_xml_set_attr_string` will resize it if the length is longer.

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: It actually already exists in `fmi3_check_variable_naming_conventions` but it's only called if FMIL is configured to check it. Unclear why we have that config option, or if there even exist more...

fmi_import_context.h
```
/**
    \brief If this configuration option is set, the model description will be
    checked to follow the variable naming conventions. Variables not following
    the convention will be logged.
*/
#define FMI_IMPORT_NAME_CHECK 1
```

```
    if (configuration & FMI3_XML_NAME_CHECK) {
        fmi3_check_variable_naming_conventions(md);
    }
```

This check is done after parsing all of modelDescription. I'll tried to move it to ModelVariables end-tag, but `configuration` variable isn't available there. Maybe not hard to fix, but don't want to do it right now. Adding FIXME instead.

Test/FMI3/fmi3_import_variable_test.cpp
```diff
@@ -361,3 +361,207 @@ TEST_CASE("Invalid structuralParameter - has dimension") {
...
+    fmi3_import_variable_t* v;
+    fmi3_import_alias_variables_t* aliases;
+    size_t nAlias;
+
+    v = fmi3_import_get_variable_by_name(fmu, baseVarName);
+    REQUIRE(v != nullptr);
+    aliases = fmi3_import_get_variable_aliases(v);
+    nAlias  = fmi3_import_get_alias_variables_number(aliases);
+    REQUIRE(nAlias == nAliasExp);
+    REQUIRE(aliases != nullptr);
```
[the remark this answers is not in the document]
filip-stenstrom: Well I thought it read easier, but OK.

Test/FMI3/fmi3_import_variable_test.cpp
```diff
@@ -361,3 +361,207 @@ TEST_CASE("Invalid structuralParameter - has dimension") {
...
+
+    return aliases;
+}
+
+void check_aliases(fmi3_import_t* fmu, const char* baseVarName, size_t nAliasExp,
```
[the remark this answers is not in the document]
filip-stenstrom: fixed

Test/FMI3/parser_test_xmls/variables/invalid/same_vr3/modelDescription.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
filip-stenstrom: Good catch, fixed

src/XML/src/FMI3/fmi3_xml_variable.c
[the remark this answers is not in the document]
filip-stenstrom: I removed the index check

src/XML/src/FMI3/fmi3_xml_variable.c
PeterMeisrimelModelon: I'm okay with fixing this later on, but does this really qualify as naming convention? 
The issue I am seeing is that a valid modelDescription should not depend on the specific variables. If we allow identical names, then functions like get_variable_by_name will yield different results depending on how the variables were ordered.

src/Import/include/FMI3/fmi3_import.h
PeterMeisrimelModelon: Ok, can always add it later on if useful/needed.

src/XML/src/FMI3/fmi3_xml_variable.c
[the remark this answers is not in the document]
PeterMeisrimelModelon: Ok, thanks for clarifying.

src/XML/src/FMI3/fmi3_xml_variable.c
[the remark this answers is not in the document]
PeterMeisrimelModelon: I suppose we should pick up this discussion at another point, since there are a bunch of these types of functions and various of them do contain the bounds checks. The behavior should be more consistent then.

src/XML/src/FMI3/fmi3_xml_variable.c
filip-stenstrom: Right that would make it ambiguous and lead to secondary errors. Same in FMI 2.0 - unclear about FMI 1.0. The config value is opt-out, so the check is currently tested in `fmi3_xml_parsing_test.c`.

I'll add to cleanup list (OneNote) and see if we can remove that config, at least for FMI 3, and clarify that it will cause issues with FMI 2.

PeterMeisrimelModelon (approved):

src/XML/src/FMI3/fmi3_xml_variable.c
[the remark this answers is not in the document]
filip-stenstrom: I agree, added cleanup (design) point about it.

PR #60, opened 2023-06-07
Enable build and test parallelization

filip-stenstrom:


.dockerignore
```diff
@@ -1,14 +0,0 @@
-# Files that will not be part of COPY to docker image
```
filip-stenstrom, line 1: File not needed since we don't do COPY any longer in Dockerfile (instead just mount).

.devcontainer/devcontainer.json
```diff
@@ -10,7 +10,8 @@
                 "vms-vscode.cpptools",
                 "ms-vscode.cmake-tools",
                 "ms-vscode.makefile-tools",
-                "ms-vscode.cpptools-extension-pack"
+                "ms-vscode.cpptools-extension-pack",
+                "matepek.vscode-catch2-test-adapter"
```
filip-stenstrom, line 14: TestMate

Config.cmake/runtime_test.cmake
```diff
@@ -152,9 +152,8 @@ include(CTest)
 set(CTEST_RETURN_SUCCESS 0)
 set(CTEST_RETURN_FAIL 1)
 
-to_native_c_path(${TEST_OUTPUT_FOLDER}/tempfolder FMU_TEMPFOLDER)
-
-file(MAKE_DIRECTORY ${FMU_TEMPFOLDER})
+to_native_c_path(${TEST_OUTPUT_FOLDER}/Temporary TEST_RESULTS_DIR)
+file(MAKE_DIRECTORY ${TEST_RESULTS_DIR})
```
filip-stenstrom, line 156: .../Temporary seems to be some directory generated by cmake, for cmake files not controlled by, especially test result files. I've changed such that we should never write to that directory.

If we need a temp directory we create it and delete it when done.

Config.cmake/test_fmi1.cmake
```diff
@@ -158,7 +153,7 @@ if(FMILIB_BUILD_BEFORE_TESTS)
         ctest_fmi_import_me_test
         ctest_fmi_import_cs_test
         ctest_fmi_import_xml_test
-        ctest_fmi_import_xml_test_empty
+        ctest_fmi_import_xml_test_brief
```
filip-stenstrom, line 156: It tests a condensed ("brief") XML. Unclear why. But it's not empty/missing at least which had me confused.

Test/FMI3/fmi3_import_sim_cs_test.c
```diff
@@ -243,7 +243,7 @@ int test_simulate_cs(fmi3_import_t * fmu)
             diff = ref_res[k] - arrValues[k];
             diff = diff > 0 ? diff : -diff;
             if (diff > tol) {
-                printf("error: incorrect final result for array idx: '%d', exp: '%f', act: '%f', diff: '%f', tol (abs.): '%f'\n", k, ref_res[k], arrValues[k], diff, tol);
+                printf("error: incorrect final result for array idx: '%zu', exp: '%f', act: '%f', diff: '%f', tol (abs.): '%f'\n", k, ref_res[k], arrValues[k], diff, tol);
```
filip-stenstrom, line 246: Just cleared a warning

PR #61, opened 2023-06-07
Allow Ninja generator (at least on linux)

filip-stenstrom:
NOTE: Currently targets dev-fs-devimp2, but will change to master before merging.


modelonrobinandersson (approved):

filip-stenstrom (commented):
Forgot to publish these self-review comments, but better late than never I guess.

PR #70, opened 2023-06-30
Initial setup for terminals and icons

PeterMeisrimelModelon:


Test/FMI1/fmi1_import_type_definitions_test.c
```diff
@@ -29,7 +29,6 @@ static int test_quantity_default(fmi1_import_t *xml)
     unsigned int n_tds;
     unsigned int i; /* td_idx */
     const char* tname;
-    const char* str_empty = "";
```
PeterMeisrimelModelon, line 32: unused.

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -263,12 +275,20 @@ void fmi3_xml_parse_free_context(fmi3_xml_parser_context_t *context) {
         jm_vector_free(jm_string)(context->attrMapById);
         context->attrMapById = 0;
     }
```
PeterMeisrimelModelon, line 277: These changes were requires to allow freeing the parser context without everything fully setup/allocated, as is the case when terminalsAndIcons does not exist.

src/XML/src/FMI3/fmi3_xml_parser.h
```diff
@@ -179,7 +180,15 @@ typedef enum fmi3_xml_attr_enu_t {
     EXPAND_XML_ELMNAME(Binary) \
     EXPAND_XML_ELMNAME(Clock) \
     EXPAND_XML_ELMNAME(String) \
-    EXPAND_XML_ELMNAME(Enumeration)
```
PeterMeisrimelModelon, line 182: The new elements could have their own list, but won't fix this for this ticket.

src/Import/include/FMI/fmi_import_util.h
```diff
@@ -53,6 +53,11 @@ FMILIB_EXPORT jm_status_enu_t fmi_import_rmdir(jm_callbacks* cb, const char* dir
 */
 FMILIB_EXPORT char* fmi_import_create_URL_from_abs_path(jm_callbacks* cb, const char* absPath);
 
+/** Given directory name fmu_unzipped_path and model identifier consturct XML file name
```
modelonrobinandersson, line 56:
```suggestion
/** Given directory name fmu_unzipped_path and the model identifier - construct the XML file name

```

src/Import/include/FMI3/fmi3_import_terminals_and_icons.h
```diff
@@ -0,0 +1,76 @@
...
+ * \brief Returns non-zero if terminalsAndIcons.xml has been found and successfully parsed.
+ * @param fmu An #fmi3_import_t object as returned by #fmi3_import_parse_xml().
+ */
+FMILIB_EXPORT int fmi3_import_get_has_terminals_and_icons(fmi3_import_t* fmu);
+
+/**
+    \brief Get terminal by terminal name.
+    @param fmu An #fmi3_import_t object as returned by #fmi3_import_parse_xml().
+    @param name terminal name
+    @return terminal pointer. NULL if terminal with given name cannot be found.
```
modelonrobinandersson, line 59:
```suggestion
    @return terminal pointer. NULL if terminal with given name does not exist.
```

src/Import/src/FMI3/fmi3_import.c
```diff
@@ -53,17 +53,26 @@ fmi3_import_t* fmi3_import_allocate(jm_callbacks* cb) {
...
     }
 
+    fmu->termIcon = fmi3_xml_allocate_terminals_and_icons(cb);
+    if (!fmu->termIcon) {
+        fmi3_xml_free_model_description(fmu->md);
+        cb->free(fmu);
+        return NULL;
+    }
+
+    jm_vector_init(char)(&fmu->logMessageBufferExpanded, 0, cb);
```
modelonrobinandersson, line 70: What does this do and/or why is it necessary? Perhaps we can add a comment above to explain it.

src/Import/src/FMI3/fmi3_import.c
```diff
@@ -116,6 +129,21 @@ fmi3_import_t* fmi3_import_parse_xml(
     }
     context->callbacks->free(xmlPath);
 
+   // Only parse terminals and icons of parsing of model description did not fail
```
modelonrobinandersson, line 132:
```suggestion
   // Only parse terminals and icons if parsing of model description did not fail
```

src/Import/src/FMI3/fmi3_import.c
```diff
@@ -116,6 +129,21 @@ fmi3_import_t* fmi3_import_parse_xml(
...
+    if (fmu) {
+        // terminalsAndIcons uses modelDescription for error checks
+        if (fmi3_xml_terminals_and_icons_set_model_description(fmu->termIcon, fmu->md)) {
+            fmi3_import_free(fmu);
+            fmu = 0;
```
modelonrobinandersson, line 137: Curious: Does it matter if we set fmu to `0` or `NULL`? I am only asking this out of consistency as it seems we set to `NULL` usually when there are errors.

src/XML/include/FMI3/fmi3_xml_terminals_and_icons.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+    \addtogroup fmi3_xml
+    @{
+    \addtogroup fmi3_xml_terminals_and_icons Functions for handling terminals and icons
+    @{
+*/
+
+typedef struct fmi3_xml_terminals_and_icons_t fmi3_xml_terminals_and_icons_t;
+typedef struct fmi3_xml_terminal_t fmi3_xml_terminal_t;
+
```
modelonrobinandersson, line 41: Would be good with a comment that explains the purpose of these.

src/XML/src/FMI3/fmi3_xml_terminals_and_icons.c
```diff
@@ -0,0 +1,238 @@
...
+
+        // free terminals
+        for (size_t i = 0; i < jm_vector_get_size(jm_voidp)(&(termIcon->terminalsOrigOrder)); i++) {
+            fmi3_xml_terminal_t* term = (fmi3_xml_terminal_t*) jm_vector_get_item(jm_voidp)(&(termIcon->terminalsOrigOrder), i);
+            fmi3_xml_free_terminal(callbacks, term);
+        }
+        jm_vector_free_data(jm_voidp)(&(termIcon->terminalsOrigOrder));
+        jm_vector_free_data(jm_named_ptr)(&termIcon->terminalsByName);
+
+        void(*cb_free)(const char*) = (void(*)(const char*))callbacks->free;
```
modelonrobinandersson, line 73: What does this do?

src/XML/src/FMI3/fmi3_xml_terminals_and_icons.c
```diff
@@ -0,0 +1,238 @@
...
+        jm_vector_qsort(jm_named_ptr)(&termIcon->terminalsByName, jm_compare_named);
+
+        // Error check, Terminal names are must be unique
+        if (nVars > 0) {  // nVars=0 would cause integer overflow in the loop condition
+            for (size_t i = 0; i < nVars-1; ++i) {
+                const char* name1 = jm_vector_get_item(jm_named_ptr)(&termIcon->terminalsByName, i).name;
+                const char* name2 = jm_vector_get_item(jm_named_ptr)(&termIcon->terminalsByName, i+1).name;
+                if(strcmp(name1, name2) == 0) {
+                    fmi3_xml_parse_fatal(context, 
+                            "Two terminals with the same name '%s' found. This is not allowed.", name1);
```
modelonrobinandersson, line 151:
```suggestion
                            "Two terminals with the same name '%s' found. This is not allowed by the specification.", name1);
```
Consider this optional.

src/XML/src/FMI3/fmi3_xml_terminals_and_icons_impl.h
```diff
@@ -0,0 +1,53 @@
...
+
+    jm_vector(char) fmi3_xml_standard_version;
+
+    jm_vector(jm_voidp) terminalsOrigOrder;
+    jm_vector(jm_named_ptr) terminalsByName;
+
+    jm_string_set names; // storage for all terminal names
+};
+
+struct fmi3_xml_terminal_t {
```
modelonrobinandersson, line 44: I would suggest adding a TODO comment here that we currently only support `name`

modelonrobinandersson (approved):
Well done, I only had minor nitpick comments that I don't consider in need of a second review.

src/Import/src/FMI3/fmi3_import.c
PeterMeisrimelModelon: I think the need to initialize buffers is straight forward, so I added comments to the `fmi3_import_t `struct that should clarify the difference between `logMessageBufferCoded `and `logMessageBufferExpanded`

src/Import/src/FMI3/fmi3_import.c
PeterMeisrimelModelon: from stddef.h:

```
#ifndef __cplusplus
#define NULL ((void *)0)
#else   /* C++ */
#define NULL 0
#endif  /* C++ */
```

So, no real practical difference, but I think it increases readability. And, yet, we are not very consistent with using it.

src/XML/src/FMI3/fmi3_xml_terminals_and_icons.c
PeterMeisrimelModelon: cb_free is a function pointer to a function with void return and const char* input.  
We assign it the callbacks->free function, with some extra typecasts, since callbacks->free is defined for jm_voidp = void* inputs.

Also added a comment for this.

src/XML/src/FMI3/fmi3_xml_terminals_and_icons_impl.h
PeterMeisrimelModelon: Added.

src/Import/src/FMI3/fmi3_import.c
filip-stenstrom: Please use NULL everywhere it is a pointer. I've tried to change it at all places I've come across it. It makes it very clear that the type is a pointer and not an int.

(I've not looked at other comments/code here, just stumbled on this thread.)

PR #73, opened 2023-07-11
Change include_directories to target_include_directories and test installation

filip-stenstrom:
Today we use include_directories and include(<e.g. fmiimport.cmake>) on all submodules. This results in that every target (add_library, add_executable) get the same include directories. This means that we won’t notice if we accidentally include headers from modules we don’t use. For example XML headers in Import headers.

What we can do instead is to use target_include_directores and then set PRIVATE or PUBLIC depending on whether we want other libraries that depend on this library to also inherit the include directories.

Later, we use target_link_libraries again with PRIVATE or PUBLIC. PRIVATE means that the included directories from the linked libraries will not be inherited further (than to the current target), while PUBLIC propagates it further. (For details read doc I suppose…)

The point is that fmilib links PRIVATE with sublibs, which means that when we build a test, if there is an XML header included in e.g. fmiimport code, then it won’t be found.

The problem still remains that we don’t know until we test the installation that all the header files (+ libs etc.) actually have been installed. So this adds a test on the installation. We want to eventually expand it to also test some FMU, e.g. by reusing some existing simulation test or so.

(The test already detected that the _dimensions_functions.h from the previous subtask was missing, but then that code was reworked.)

The code is put in a separate CMake project to avoid inheriting anything from global definitions in the FMIL CMake project, and I’m also not certain one can call ‘install’ from within CMake in a convenient way.


Test/test_installation/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+elseif(MINGW)
+    set_property(TARGET fmilib_shared PROPERTY IMPORTED_LOCATION "${FMIL_HOME}/lib/libfmilib_shared.dll")
+    set_property(TARGET fmilib_shared PROPERTY IMPORTED_IMPLIB   "${FMIL_HOME}/lib/libfmilib_shared.dll.a")
+else()
+    set_property(TARGET fmilib_shared PROPERTY IMPORTED_LOCATION "${FMIL_HOME}/lib/libfmilib_shared.so")
+endif()
+
+# Build the test
+add_executable(main_shared main.c)
+target_link_libraries(main_shared PRIVATE fmilib_shared)
```
PeterMeisrimelModelon, line 94: Isn't this line covered by both of the branches in the following if?

Config.cmake/runtime_test.cmake
```diff
@@ -12,40 +12,65 @@
...
+# Only the directories that are specific to tests.
+set(FMIL_TEST_INCLUDE_DIRS
+    ${FMILIB_TESTCONFIG_INCLUDE_DIR}
+    ${FMIL_TEST_DIR} 
+    ${FMIL_TEST_DIR}/FMI1
+    ${FMIL_TEST_DIR}/FMI2
+    ${FMIL_TEST_DIR}/FMI3
 )
 
+# Workaround to allow testing with sublibs on MinGW (unclear why it works with MSVC).
```
PeterMeisrimelModelon, line 26: Should we mark this as XXX, FIXME or similar for future reference?

PeterMeisrimelModelon (approved):

Test/test_installation/CMakeLists.txt
filip-stenstrom: You're right, and same for static fmilib, will fix

Makefile
```diff
@@ -13,7 +13,7 @@ include $(CONFIG_FILE)
 
 install: generate
 	cd $(BUILD_DIR) && \
-		cmake --build . --parallel $(NPROC) --config $(BUILD_TYPE) --target '$@'
+		cmake --build . --parallel $(NPROC) --config $(BUILD_TYPE) --target install
```
filip-stenstrom: Refactor for readability

.devcontainer/centos/devcontainer.json
```diff
@@ -1,6 +1,7 @@
 {
+    "name": "centos",
     "build": {
-        "dockerfile": "../build/docker/Dockerfile_ubuntu"
+        "dockerfile": "${localWorkspaceFolder}/build/docker/Dockerfile_centos"
```
filip-stenstrom, line 4: Needed centos devcontainer to be able to debug

build/config/ubuntu64
```diff
@@ -31,4 +31,5 @@ TEST_LOCALE=1
 
 # Path used in container for the mounted directories when building via docker
 MOUNTED_DIR=/mnt/fmi-library
-DOCKERFILE_NAME=Dockerfile_ubuntu
+DOCKERFILE_NAME=$(CONFIG_NAME).Dockerfile
```
filip-stenstrom, line 34: Rename for readability.

build/docker/Makefile
```diff
@@ -4,17 +4,15 @@ include $(CONFIG_FILE)
 #
 # NOTE: Working directory when calling this file must be project root.
 
-IMG_TAG=fmil:test
-
 .PHONY: docker_image install test documentation docker_run_debug
 
 docker_image:
-	docker build -f build/docker/$(DOCKERFILE_NAME) -t $(IMG_TAG) .
+	docker build -f build/docker/$(DOCKERFILE_NAME) -t $(DOCKER_IMAGE_TAG) .
```
filip-stenstrom, line 10: Here's the main change. Before the tag name was `fmil:test` for both centos and ubuntu. When the jobs run in parallel on the same machine, this causes a race condition. Now the name will unique for each platform.

Here is still a race condition. Let's say I make a change in the image on my dev branch and start a build at same time as some other job is building. Then they will have the same name but different image content, so same issue. The way to get around it would be to have a separate build to create the image like we do in other projects, but I don't feel it's worth the effort right now. (This in general applies for any Jenkins job which builds a docker image as part of CI testing.)

PeterMeisrimelModelon (approved):
Looks great!

PR #81, opened 2023-07-19
Fixed Dockerfile reference for devcontainers

PeterMeisrimelModelon:


.vscode/settings.json
```diff
@@ -2,7 +2,7 @@
     "cmake.buildDirectory": "${workspaceFolder}/build_cmake",  // The default ('build') is already used
     "cmake.configureOnOpen": false,
     "cmake.preferredGenerators": [
-        "Ninja",           // NOTE: Jenkins still runs with "Unix Makefiles" since other downstream projects use that
+        // "Ninja",           // NOTE: Jenkins still runs with "Unix Makefiles" since other downstream projects use that
```
PeterMeisrimelModelon, line 5: Not sure what the appropriate action here is, seeing that this doesn't seem to work by default. May need some more looking into it.

modelonrobinandersson (approved):

.vscode/settings.json
modelonrobinandersson: I would suggest to leave this for now and maybe we can talk about it again in the end of August. At least it is working now, well done!

PR #90, opened 2023-10-19
Completely disable jm_locale_test if locale testing is disabled

### LINKED ISSUE #80: Undefined symbol in MinGW DEBUG build with 3.0a2
Build 3.0a2 as a _MinGW Makefiles_ project with
````bat
cmake -G "MinGW Makefiles" -DCMAKE_BUILD_TYPE=Debug ../fmi-library-3.0a2
cmake --build .
````
results in an error when linking `jm_locale_test`:
````
[....]
[ 55%] Linking C executable jm_locale_test.exe
C:/..../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\jm_locale_test.dir/objects.a(jm_locale_test.c.obj): in function `test_parse_with_locale':
C:/..../fmi-library-build/fmi-library-3.0a2/Test/jm_locale_test.c:39: undefined reference to `__imp_jm_get_default_callbacks'
collect2.exe: error: ld returned 1 exit status
````

Building with `-DCMAKE_BUILD_TYPE=Release` completes without errors. The 3.0a1 _DEBUG_ build completes without errors, as does skipping the tests with `-DFMILIB_BUILD_TESTS=0`.

````
cmake version 3.26.4
gcc (MinGW-W64 x86_64-ucrt-mcf-seh, built by Brecht Sanders) 13.1.0
````

### PR DESCRIPTION
**Pull request by filip-stenstrom**
Attemps to fix #80 . I didn't manage to reproduce it locally, but I hope this will fix it. Problem summary:

```
C:/..../fmi-library-build/fmi-library-3.0a2/Test/jm_locale_test.c:39: undefined reference to `__imp_jm_get_default_callbacks'
```

You get this when you expect to link with an import lib. See comment in code:
image

`jm_locale_test` only depends on `jmutils`, which is a static lib, so we must always set `FMIL_LINK_WITH_SUBLIBS`. We didn't do this unless `FMILIB_TEST_LOCALE` was on, which is default off. We however still built the test itself, which had functions that referenced the import lib function. I changed so we don't even build the test unless that option is on.

### DIFF
--- a/Config.cmake/runtime_test.cmake
+++ b/Config.cmake/runtime_test.cmake
@@ -66,18 +66,14 @@ if(FMILIB_TEST_LOCALE)
     target_link_libraries(jm_locale_test jmutils)
     target_include_directories(jm_locale_test PRIVATE ${FMIL_TEST_INCLUDE_DIRS})
     target_compile_definitions(jm_locale_test PRIVATE ${FMIL_LINK_WITH_SUBLIBS})
+    add_test(ctest_jm_locale_test jm_locale_test)
+    set_tests_properties(ctest_jm_locale_test PROPERTIES DEPENDS ctest_build_all)
 endif()
 
 #Create function that zips the dummy FMUs
 add_executable(compress_test_fmu_zip ${FMIL_TEST_DIR}/compress_test_fmu_zip.c)
 target_link_libraries(compress_test_fmu_zip fmizip)
 target_include_directories(compress_test_fmu_zip PRIVATE ${FMIL_TEST_INCLUDE_DIRS})
-
-set_target_properties(
-    jm_vector_test
-    jm_locale_test
-    compress_test_fmu_zip
-    PROPERTIES FOLDER "Test")
 
 #Path to the executable
 get_property(COMPRESS_EXECUTABLE TARGET compress_test_fmu_zip PROPERTY LOCATION)
@@ -190,8 +186,6 @@ if(FMILIB_BUILD_BEFORE_TESTS)
         COMMAND "${CMAKE_COMMAND}" --build ${FMILIBRARYBUILD} --config $<CONFIGURATION>)
 endif()
 
-add_test(ctest_jm_locale_test jm_locale_test)
-
 add_test(ctest_fmi_zip_unzip_test fmi_zip_unzip_test)
 add_test(ctest_fmi_zip_zip_test fmi_zip_zip_test)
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Config.cmake/runtime_test.cmake [hunk 1]
```diff
@@ -61,24 +61,20 @@ target_link_libraries(jm_vector_test PRIVATE jmutils)
...
 add_executable(compress_test_fmu_zip ${FMIL_TEST_DIR}/compress_test_fmu_zip.c)
 target_link_libraries(compress_test_fmu_zip fmizip)
 target_include_directories(compress_test_fmu_zip PRIVATE ${FMIL_TEST_INCLUDE_DIRS})
 
-set_target_properties(
-    jm_vector_test
-    jm_locale_test
-    compress_test_fmu_zip
-    PROPERTIES FOLDER "Test")
-
```
filip-stenstrom, line 81: This is only for developing on Visual Studio IDE. We're not consistent with `FOLDER` regardless, so I just removed it. As we're moving to `add_catch2_test` it will be handled automatically.

Config.cmake/runtime_test.cmake
```diff
@@ -61,24 +61,20 @@ target_link_libraries(jm_vector_test PRIVATE jmutils)
...
 # Test: jm locale
-add_executable(jm_locale_test ${FMIL_TEST_DIR}/jm_locale_test.c)
-target_link_libraries(jm_locale_test jmutils)
-target_include_directories(jm_locale_test PRIVATE ${FMIL_TEST_INCLUDE_DIRS})
 if(FMILIB_TEST_LOCALE)
-    target_compile_definitions(jm_locale_test PRIVATE -DFMILIB_TEST_LOCALE ${FMIL_LINK_WITH_SUBLIBS})
+    add_executable(jm_locale_test ${FMIL_TEST_DIR}/jm_locale_test.c)
+    target_link_libraries(jm_locale_test jmutils)
+    target_include_directories(jm_locale_test PRIVATE ${FMIL_TEST_INCLUDE_DIRS})
+    target_compile_definitions(jm_locale_test PRIVATE ${FMIL_LINK_WITH_SUBLIBS})
```
PeterMeisrimelModelon, line 68: This one seems to be redundant with the test change?

filip-stenstrom: I don't think so. We still make use of `jm_get_default_callbacks` which is defined in the header as:

```
FMILIB_EXPORT
jm_callbacks* jm_get_default_callbacks(void);
```

And FMILIB_EXPORT will expand to `__declspec(dllimport)` unless we set that flag, which will cause the issue with it expecting to link with an import lib instead of a static lib (see `config_fmilib.h.cmake` for details on how FMILIB_EXPORT is expanded).

PeterMeisrimelModelon: Ah ok, thanks for the clarification, I somehow missed that link.

PR #97, opened 2023-11-13
Make FMILibrary compatible for cross compilation

### LINKED ISSUE #96: Cross compilation compatible
We use FMI Library for our openPASS and one of our goals is to provide executable under windows. Therefore, it would be great if source would be cross compilation compatible. So that we can cross compilation FMILibrary with ubuntu as host system and MINGW as target system.

We have observed few things, when cross compiling FMILibrary.
1. CMake Compiler and System variables are not forwarded when build expat library
2. "Shlwapi" is case sensitive when building on linux.

raghunandannm:
resolves #96


modelonrobinandersson:
Hi @raghunandannm, for contributing to FMI Library see the section in the README https://github.com/modelon-community/fmi-library#fmi-library

PR #98, opened 2023-11-14
Update fmi2_enums.c to solve independent (time) variable start attrit…

lionelbelmon:
…ube error

the xml check fails on independent variable , it asks for start attribute, (it should not).

this seemed to be working before changing this enum.

either change the check funnction (another pull request) or


PeterMeisrimelModelon:
I think this issue should be addressed with https://github.com/modelon-community/fmi-library/pull/129 ?

Issue #110, opened 2024-02-21
Unable to load FMU3 exported by Dyomla


Thantalos:
fmi-library fails to load FMU3s exported by Dymola 2023x or 2024x. It first reports a number of problems and then ultimately fails to load. The 2024x version produces more output than the 2023 version. Here I have an example FMU (it's a simple model, the PID controller example from the standard library):
[dymola2024x.zip](https://github.com/modelon-community/fmi-library/files/14356797/dymola2024x.zip)
And here the output produced when loading:
```
[2024-02-21 09:48:36.421] [FMILIB] [info] XML specifies FMI standard version 3.0
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:178] Unknown element 'Annotation' in XML, skipping
[2024-02-21 09:48:36.422] [FMI3XML] [warning] [Line:180] Skipping nested XML element 'DeclarationOrder'
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:194] Unknown element 'Annotation' in XML, skipping
[2024-02-21 09:48:36.422] [FMI3XML] [warning] [Line:196] Skipping nested XML element 'DeclarationOrder'
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:208] Unknown element 'Annotation' in XML, skipping
[2024-02-21 09:48:36.422] [FMI3XML] [warning] [Line:210] Skipping nested XML element 'DeclarationOrder'
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:233] Unknown element 'Annotation' in XML, skipping
[2024-02-21 09:48:36.422] [FMI3XML] [warning] [Line:235] Skipping nested XML element 'DeclarationOrder'
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:248] Unknown element 'Annotation' in XML, skipping
[2024-02-21 09:48:36.422] [FMI3XML] [warning] [Line:250] Skipping nested XML element 'DeclarationOrder'
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:264] Unknown element 'Annotation' in XML, skipping
[2024-02-21 09:48:36.422] [FMI3XML] [warning] [Line:266] Skipping nested XML element 'DeclarationOrder'
[2024-02-21 09:48:36.422] [FMI3XML] [error] [Line:280] Unknown element 'Annotation' in XML, skipping
[...]

PR #112, opened 2024-02-26
Fix for dependencies with very large VRs

PeterMeisrimelModelon:
https://github.com/modelon-community/fmi-library/issues/110

### DIFF
--- a/src/XML/src/FMI3/fmi3_xml_model_structure.c
+++ b/src/XML/src/FMI3/fmi3_xml_model_structure.c
@@ -253,7 +253,7 @@ static int fmi3_xml_parse_dependencies(fmi3_xml_parser_context_t* context,
     }
     if (listInd) {
         const char* cur = listInd;
-        int ind;
+        long ind;
         while (*cur) {
             char ch = *cur;
             while ((ch ==' ') || (ch == '\t') || (ch =='\n') || (ch == '\r')) {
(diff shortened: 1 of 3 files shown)


src/XML/src/FMI3/fmi3_xml_model_structure.c
```diff
@@ -253,21 +253,21 @@ static int fmi3_xml_parse_dependencies(fmi3_xml_parser_context_t* context,
     }
     if (listInd) {
         const char* cur = listInd;
-        int ind;
+        long ind;
```
PeterMeisrimelModelon, line 256: While formally an unsigned int, we would no longer be able to catch e.g., negative values here.

src/XML/src/FMI3/fmi3_xml_model_structure.c
```diff
@@ -253,21 +253,21 @@ static int fmi3_xml_parse_dependencies(fmi3_xml_parser_context_t* context,
     }
     if (listInd) {
         const char* cur = listInd;
-        int ind;
+        long long ind; // at least twice the size of unsigned int32
```
PeterMeisrimelModelon, line 256: `long long` is required to catch e.g., negative values here.

modelonrobinandersson (approved):

Thantalos:
thank you

PR #113, opened 2024-02-26
CMake: Unset CMAKE_INSTALL_PREFIX

### LINKED ISSUE #12: Overriding CMAKE_INSTALL_PREFIX is bad practice
https://github.com/modelon-community/fmi-library/blob/998768cddacce99727d633d045fa6d25314e1d9a/CMakeLists.txt#L88

Though calling it "practice" implies that I've ever seen it done elsewhere. This is silently and forcibly(!) setting a core cmake option to something other than what the user might specify, which is not great. In isolation to one project (this one), it's... probably fine? When building code with/against FMILibrary it blows up.

We are attempting to use FMILibrary as a sub-project, and this line nukes the install path for the rest of our code, unless we set the FMI-specific install path to the same one. Which, to be fair, we _probably_ would, but it's a pit that we easily fell into.

My recommendation would be to simply use CMAKE_INSTALL_PREFIX natively, but at the very least don't override user settings for standard cmake variables.

jschueller:
Closes #12


stefansli:
@PeterMeisrimelModelon @modelonrobinandersson  @MikaelJohnssonModelon  please asap. It's really cumbersome to include fmilibrary in any package management system. If the community tries to fix the issues and creates pull requests it's very sad to see those pull requests gett no attendion for **months**.

modelonrobinandersson:
Hi @stefansli and @jschueller.
I just opened #126 to include 113, 114 and some other updates which simplifies this a bit. As soon as I get all the tests to pass on Linux, Windows (MinGW and MSVC (MD/MT)) I will merge it.

I've also seen the other PRs but I will have to take them later as they cause some issues with older compilers on Windows.

PR #114, opened 2024-02-26
CMake: Use GNUInstallDirs

jschueller:
Install files in standard location and allows to override each setting


CMakeLists.txt
```diff
@@ -385,21 +385,23 @@ if(FMILIB_BUILD_SHARED_LIB)
...
 
+include (GNUInstallDirs)
+
 install(TARGETS ${FMILIB_TARGETS}
-    ARCHIVE DESTINATION lib
-    LIBRARY DESTINATION lib
-    RUNTIME DESTINATION lib
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
```
modelonrobinandersson, line 393: @jschueller I tried this PR but for some reason it doesn't work with any Windows builds (MinGW or msvc), for Windows. I haven't had time to look into the details but any ideas on why? If I have `RUNTIME DESTINATION ${CMAKE_INSTALL_LIBDIR}` it works without issues for all platforms.

jschueller: maybe  your scripts are expecting the dll to be installed in PREFIX/lib (eg PATH variable ?), but now dlls are installed to PREFIX/bin instead and cannot be loaded

modelonrobinandersson: I see. I think I have to get into contact with the author of that change and see if he knows anything, however it will likely be in a few weeks due to vacations. Will this be unusable for you if I would merge the `GNUInstallDirs` change with `${CMAKE_INSTALL_BINDIR}` on all 3 above?

jschueller: you mean LIBDIR ? then it wouldnt change much, better wait then

modelonrobinandersson: Correct, I will see if I have some time to get it working this week or next. Thanks again for your input.

modelonrobinandersson: @jschueller I have fixed the issue now, I will close this and see if I can get it merged tomorrow.

PR #116, opened 2024-02-26
CI: Add GHA script

jschueller:


filip-stenstrom (changes requested):
Overall good and will be useful for fast feedback, especially for contributors. See comments.

jschueller:
done

filip-stenstrom (approved):

PR #129, opened 2024-07-08
Correct error check on causality="independent" and start attribute(s)

PeterMeisrimelModelon:


src/XML/src/FMI2/fmi2_xml_variable.c
```diff
@@ -479,8 +479,8 @@ int fmi2_xml_handle_ScalarVariable(fmi2_xml_parser_context_t *context, const cha
 
 int fmi2_xml_get_has_start(fmi2_xml_parser_context_t *context, fmi2_xml_variable_t* variable) {
     int hasStart = fmi2_xml_is_attr_defined(context, fmi_attr_id_start);
-    if(!hasStart)  {
-        if (variable->initial != (char)fmi2_initial_enu_calculated) {
+    if (!hasStart) {
```
PeterMeisrimelModelon, line 482: specification: "If one variable is defined as "independent", it must be defined as "Real" without a "start" attribute."

modelonrobinandersson: Doesn't this imply that the test-xml file above is invalid since it is not of type `Real`?
I.e. `<Int64 name="i64_independent" valueReference="17" causality="independent"/>`

I couldn't find the exact reference you quoted above but from FMI 3.0:
```
The independent variable must be defined as a floating point type without a [start](https://fmi-standard.org/docs/3.0/#start) attribute. It is not allowed to call function [fmi3Set{VariableType}](https://fmi-standard.org/docs/3.0/#get-and-set-variable-values) on an [independent](https://fmi-standard.org/docs/3.0/#independent) variable.
```

PeterMeisrimelModelon: Correct, yet another error check that we didn't have. Added it, see latest 2 commits.

In FMI2 this will be caught by invalid combination of causality + variability, which set the variability to continuous and then fails due to having a continuous e.g., Integer. Not optimal, but don't think it is worth working on.

modelonrobinandersson: Looks good, I think just a changelog entry is missing but other than that great work!

modelonrobinandersson (approved):

src/XML/src/FMI2/fmi2_xml_variable.c
PeterMeisrimelModelon: Added one for the fixed incorrect error check; I don't think the new error checks for FMI3 are worth highlighting.

Issue #100, opened 2023-11-21
CMake option to not use bundled thirdparty


joakimono:
We are working on a so-called conan recipe to build fmi-library on conan center index (CCI), so that it can be made available via [conan](https://conan.io/) (a C / C++ package manager). 

The recipe will be maintained by the open source community. A good practice on CCI is to let conan handle the third-party dependencies instead of building third-party code. Currently (for 2.4.1), this necessitates several patches to workaround the external project compilation, see https://github.com/conan-io/conan-center-index/pull/20256, https://github.com/joakimono/conan-center-index/tree/fmilibrary/new-package/recipes/fmilib/all/patches.

It would be more maintainable if fmi-library could optionally disable bundled third-party dependencies, e.g. through a CMake option variable.

Could you please consider such feature on your next major release?

modelonrobinandersson:
Hi @joakimono, thank you for the request. It sounds like a good idea and I'll see what we can do about this.

jschueller:
I also added a new option to use all system libraries

PR #134, opened 2024-07-28
build: Add CMake option to use external expat

uilianries:
Hello all!

This PR adds the CMake option `FMILIB_SYSTEM_EXPAT`, to be able to consume expat from system, provided by a package manager.

The official CMake module for Expat is [FindEXPAT](https://cmake.org/cmake/help/latest/module/FindEXPAT.html) and provides both `EXPAT_INCLUDE_DIRS` and `EXPAT_LIBRARIES`.

By default, this option is disabled, so it should not break users.

The entire build log using the system expat is here:

[fmilib-system-expat-build.log](https://github.com/user-attachments/files/16404576/fmilib-system-expat-build.log)

It's possible to see:

> -- Found EXPAT: /usr/lib/x86_64-linux-gnu/libexpat.so (found version "2.1.0")

Related to #100

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,7 +62,7 @@ option(FMILIB_BUILD_STATIC_LIB "Build the library as static." ON)
 option(FMILIB_BUILD_SHARED_LIB "Build the library as shared (dll/so/dylib)." ON)
 option(FMILIB_INSTALL_SUBLIBS  "Install sub-libraries (fmicapi. fmixml, jmutil, fmiimport, fmizip)" OFF)
 option(FMILIB_FIND_PACKAGE_ZLIB  "Use find_package for zlib, instead of version in FMILibrary/Thirdparty." OFF)
-option(FMILIB_SYSTEM_EXPAT "Use system expat library instead of the one in FMILibrary/Thirdparty." OFF)
+option(FMILIB_SYSTEM_EXPAT "Use system expat library instead of the one in FMILibrary/Thirdparty. Static builds of FMI Library will not bundle expat." OFF)
 mark_as_advanced(FMILIB_FIND_PACKAGE_ZLIB)
 
 if(NOT (WIN32 OR CYGWIN OR APPLE))
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,7 @@ option(FMILIB_BUILD_STATIC_LIB "Build the library as static." ON)
 option(FMILIB_BUILD_SHARED_LIB "Build the library as shared (dll/so/dylib)." ON)
 option(FMILIB_INSTALL_SUBLIBS  "Install sub-libraries (fmicapi. fmixml, jmutil, fmiimport, fmizip)" OFF)
 option(FMILIB_FIND_PACKAGE_ZLIB  "Use find_package for zlib, instead of version in FMILibrary/Thirdparty." OFF)
+option(FMILIB_SYSTEM_EXPAT "Use system expat library instead of the one in FMILibrary/Thirdparty." OFF)
 mark_as_advanced(FMILIB_FIND_PACKAGE_ZLIB)
 
 if(NOT (WIN32 OR CYGWIN OR APPLE))
@@ -288,7 +289,10 @@ if(FMILIB_BUILD_TESTS)
     )
 endif()
 
-set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi expat minizip zlib c99snprintf)
+set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi minizip zlib c99snprintf)
+if (NOT FMILIB_SYSTEM_EXPAT)
+    list(APPEND FMILIB_SHARED_SUBLIBS expat)
+endif()
 set(FMILIB_SUBLIBS fmiimport jmutils ${FMILIB_SHARED_SUBLIBS})
 # XXX: Unclear why we create intermediate sublibs instead of just giving everything as source.
 # Probably also wouldn't get problems with using PRIVATE/PUBLIC on linking sublibs on Windows.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,7 +290,7 @@ if(FMILIB_BUILD_TESTS)
 endif()
 
 set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi minizip zlib c99snprintf)
-if (NOT FMILIB_SYSTEM_EXPAT)
+if(NOT FMILIB_SYSTEM_EXPAT)
     list(APPEND FMILIB_SHARED_SUBLIBS expat)
 endif()
 set(FMILIB_SUBLIBS fmiimport jmutils ${FMILIB_SHARED_SUBLIBS})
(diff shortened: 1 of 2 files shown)


modelonrobinandersson:
Hi @uilianries and thank you for your PR. There is just something I have to check on my end, I will get back to you on Monday again.

uilianries:
@modelonrobinandersson Thank you! Please, take your time to validate this change, I don't want to break users :sweat_smile:

modelonrobinandersson:
Hi again @uilianries, in order to accept contributions to FMIL there is the following requirement: https://github.com/modelon-community/contributing
I will only be able to merge your request after this process has been finished, let me know if you have any questions.

uilianries:
@modelonrobinandersson Hello, is it possible to send electronic signed instead? I don't have a printer.

modelonrobinandersson:
> @modelonrobinandersson Hello, is it possible to send electronic signed instead? I don't have a printer.

Hi! Yes that works as well.

uilianries:
@modelonrobinandersson Hello again, I just sent the email for the CLA, but there is not technical issue on Modelon side. I opened a separated issue reporting the case: https://github.com/modelon-community/contributing/issues/2

filip-stenstrom:
To be clear, I will review this, but need to finish some other work first. Will look at this when I can.

CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ option(FMILIB_BUILD_STATIC_LIB "Build the library as static." ON)
 option(FMILIB_BUILD_SHARED_LIB "Build the library as shared (dll/so/dylib)." ON)
 option(FMILIB_INSTALL_SUBLIBS  "Install sub-libraries (fmicapi. fmixml, jmutil, fmiimport, fmizip)" OFF)
 option(FMILIB_FIND_PACKAGE_ZLIB  "Use find_package for zlib, instead of version in FMILibrary/Thirdparty." OFF)
+option(FMILIB_SYSTEM_EXPAT "Use system expat library instead of the one in FMILibrary/Thirdparty." OFF)
```
filip-stenstrom, line 65:
```suggestion
option(FMILIB_SYSTEM_EXPAT "Use system expat library instead of the one in FMILibrary/Thirdparty. Static builds of FMI Library will not bundle expat." OFF)
```

CMakeLists.txt
```diff
@@ -288,7 +289,10 @@ if(FMILIB_BUILD_TESTS)
     )
 endif()
 
-set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi expat minizip zlib c99snprintf)
+set(FMILIB_SHARED_SUBLIBS fmixml fmizip fmicapi minizip zlib c99snprintf)
+if (NOT FMILIB_SYSTEM_EXPAT)
```
filip-stenstrom, line 293:
```suggestion
if(NOT FMILIB_SYSTEM_EXPAT)
```

Config.cmake/fmixml.cmake
```diff
@@ -179,92 +179,97 @@ set(FMIXMLSOURCE
...
+        -DCMAKE_EXE_LINKER_FLAGS:STRING=${CMAKE_EXE_LINKER_FLAGS}
+        -DCMAKE_LINK_LIBRARY_FLAG:STRING=${CMAKE_LINK_LIBRARY_FLAG}
+        -DCMAKE_MODULE_LINKER_FLAGS:STRING=${CMAKE_MODULE_LINKER_FLAGS}
+        -DCMAKE_SHARED_LINKER_FLAGS:STRING=${CMAKE_SHARED_LINKER_FLAGS}
+        -DCMAKE_INSTALL_PREFIX:PATH=${CMAKE_BINARY_DIR}/ExpatEx/install
+    )
+
+    ExternalProject_Add(
+        expatex
+        PREFIX "${FMIXML_EXPAT_DIR}"
```
filip-stenstrom, line 214: The declaration of FMIXML_EXPAT_DIR could be moved to inside this `else` branch, e.g. a few lines up.

Config.cmake/fmixml.cmake
```diff
@@ -179,92 +179,97 @@ set(FMIXMLSOURCE
...
-    set(EXPAT_LIB_PREFIX ${CMAKE_STATIC_LIBRARY_PREFIX})
-endif()
-
-if("${CMAKE_CFG_INTDIR}" STREQUAL ".")
-    # Ninja complains about 'ExpatEx/./libexpat.a' otherwise. Probably because
-    # generator expressions in mergestaticlibs give slighlty different paths.
-    set(expatlib "${CMAKE_BINARY_DIR}/ExpatEx/${EXPAT_LIB_PREFIX}expat${CMAKE_STATIC_LIBRARY_SUFFIX}")
+if(FMILIB_SYSTEM_EXPAT)
+    find_package(EXPAT REQUIRED)
+    set(expat ${EXPAT_LIBRARIES})
```
filip-stenstrom, line 184:
```suggestion
```

Not used. Only place `expat` is used is here:

```
target_link_libraries(fmixml
    PRIVATE expat
    PUBLIC  jmutils
)
```

However this is pointing to the `add_library` target `expat`, not the variable `expat`. Ideally we'd probably want `EXPAT::EXPAT` to be used there, but it seems the find_package call creates the `expat` library, so not needed for now.

Config.cmake/fmixml.cmake
```diff
@@ -179,92 +179,97 @@ set(FMIXMLSOURCE
...
-else()
-    set(EXPAT_LIB_PREFIX ${CMAKE_STATIC_LIBRARY_PREFIX})
-endif()
-
-if("${CMAKE_CFG_INTDIR}" STREQUAL ".")
-    # Ninja complains about 'ExpatEx/./libexpat.a' otherwise. Probably because
-    # generator expressions in mergestaticlibs give slighlty different paths.
-    set(expatlib "${CMAKE_BINARY_DIR}/ExpatEx/${EXPAT_LIB_PREFIX}expat${CMAKE_STATIC_LIBRARY_SUFFIX}")
+if(FMILIB_SYSTEM_EXPAT)
+    find_package(EXPAT REQUIRED)
```
filip-stenstrom, line 183:
```suggestion
    # On success defines (at least) the following variables:
    #   - EXPAT_INCLUDE_DIRS
    #   - EXPAT_LIBRARIES
    # And the following target:
    #   - expat
    find_package(EXPAT REQUIRED)
```

Just to make it easier to find where implicitly defined variables comes from.

filip-stenstrom (changes requested):
Good addition. FMIL should have better dependency management and this is one step in that direction.

Things were made a bit complicated since we already have e.g. FMILIB_FIND_PACKAGE_ZLIB which behaves slightly different (bundles zlib with FMIL). But we think not doing that (like done here), and instead listing dependencies is the way forward.

I've added suggestions for most changes.

Thanks for contributing, and sorry for delay in reviewing.

Config.cmake/fmixml.cmake
uilianries: Solved by https://github.com/modelon-community/fmi-library/pull/134/commits/84d61f36524d9b29e318e51acbafe6b2216a1cfb

uilianries:
@filip-stenstrom Thank you for your review. All committed all your suggestions! Please, review again when possible. Don't worry about the delay, as maintainer of open source project too, I totally understand :wink:

jschueller:
hi,  do you plan to do zlib/minizip too ?

uilianries:
@jschueller That's possible.

jschueller:
that would be great!

PR #137, opened 2024-08-08
Refactoring FMI3 XML module; re-arranging code into more sensible file structures

PeterMeisrimelModelon:
Refactoring diff is naturally a mess, should be manageable by looking at it commit by commit though.

Summary:

1. Moved scheme specific code to fmi(3)_xml_*_scheme.h/c files
2. Moved common scheme code to a *_scheme_base file; to avoid circular includes.
3. General scheme interaction code is now in fmi3_xml_parser_scheme.h/c, this includes some code previously in fmi3_xml_parser.c
4. Migrated headers to /include
5. Cleaned up some includes


modelonrobinandersson (approved):
Nice refactoring, I agree I think it makes more sense

PR #144, opened 2024-10-04
Removed chmod of script + added file separator for resourcePath

modelonrobinandersson:
https://fmi-standard.org/docs/3.0.1/#resourcePath

### DIFF
--- a/src/Import/src/FMI3/fmi3_import.c
+++ b/src/Import/src/FMI3/fmi3_import.c
@@ -110,7 +110,7 @@ fmi3_import_t* fmi3_import_parse_xml(
 
     if (jm_get_dir_abspath(context->callbacks, dirPath, absPath, FILENAME_MAX + 2)) {
         size_t len = strlen(absPath);
-        strcpy(absPath + len, FMI_FILE_SEP "resources");
+        strcpy(absPath + len, FMI_FILE_SEP "resources" FMI_FILE_SEP);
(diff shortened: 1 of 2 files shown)


src/Import/src/FMI3/fmi3_import.c
```diff
@@ -110,7 +110,7 @@ fmi3_import_t* fmi3_import_parse_xml(
 
     if (jm_get_dir_abspath(context->callbacks, dirPath, absPath, FILENAME_MAX + 2)) {
         size_t len = strlen(absPath);
-        strcpy(absPath + len, FMI_FILE_SEP "resources");
+        strcpy(absPath + len, FMI_FILE_SEP "resources" FMI_FILE_SEP);
```
filip-stenstrom, line 113: Would probably be a good idea to add a comment here that FMI 3 requires trailing / but FMI 2 doesn't.

filip-stenstrom (commented):
- Maybe write something in release notes.
- Optional, but if you want to add a test, you can probably add an assert fmu3_model.c instantiation for the dummy FMU that resourcePath ends with /.

src/Import/src/FMI3/fmi3_import.c
modelonrobinandersson: Good idea, fixed!

filip-stenstrom (approved):

PR #149, opened 2024-10-28
Use plain path instead of URI for resourcePath

filip-stenstrom:
FMI 2 expects a URI for the `resourcePath` argument given to fmi2Instantiate, but in FMI 3 it's just a plain path. I.e. not a URI, and should therefore not get `file` URI prefix.

### DIFF
--- a/Test/FMI3/fmi3_import_sim_me_test.cpp
+++ b/Test/FMI3/fmi3_import_sim_me_test.cpp
@@ -356,6 +356,9 @@ TEST_CASE("Test fallback/default value(s) passed to fmi3Instantiate") {
     fmi3_import_t* fmu = tfmu->fmu;
     REQUIRE(fmu != nullptr);
 
+    // TODO: Various issues related to parsing of Annotations
+    REQUIRE(fmi3_testutil_get_num_problems(tfmu) == 5);
+
     REQUIRE((fmi3_import_get_fmu_kind(fmu) & fmi3_fmu_kind_me) == fmi3_fmu_kind_me);
 
     // Set logger:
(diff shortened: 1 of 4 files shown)


Test/FMI3/fmi3_import_sim_me_test.cpp
```diff
@@ -311,3 +311,79 @@ TEST_CASE("Model-Exchange FMU example") {
...
+    // Main part of this test:
+    char expected[4096];
+    strcpy(expected, tmpPath);
+#ifdef WIN32
+    strcpy(expected + strlen(expected), "\\resources\\");
+#else
+    strcpy(expected + strlen(expected), "/resources/");
+#endif
+    REQUIRE(strcmp(expected, loggedResourcePath) == 0);
+
```
PeterMeisrimelModelon, line 381: A while ago we changed most/all tests to use `fmi3_testutil_parse_xml_with_log(...)`, such that we can use `fmi3_testutil_get_num_problems(...)` to have tighter control/regression checks on warnings or other issues. See e.g., the test above. 

It would be nice to have that here as well.

PeterMeisrimelModelon (approved):
Looks great, minor comment on the test.

Test/FMI3/fmi3_import_sim_me_test.cpp
filip-stenstrom: Fixed. (Initially I had the custom logger used in this test for XML testing also, before I realized I could have different logger for CAPI calls.)

PeterMeisrimelModelon: Can you add an appropriate `REQUIRE(fmi3_testutil_get_num_problems(tfmu) == <num>);` at the end of the test, so we have the regression check on any issues? + If `<num>` is not `0` and acceptable here (since the dummy FMUs contain some issues), a `TODO` comment on the issue(s)?

Feel free to merge after that!

filip-stenstrom: This test hasn't really anything to do with XML, but I've added it. Merging.

PR #152, opened 2024-10-31
Added fix for improved CMake version compliance

modelonrobinandersson:


filip-stenstrom (approved):
OK.

Ideally we should update CMake minimum required but seems there are some issues with that: https://github.com/modelon-community/fmi-library/pull/115

And we also want to make a release soon, so this quick fix is OK for me.

PR #166, opened 2025-04-16
Fixing interval&shift decimal functions in accordance to fix in speci…

PeterMeisrimelModelon:
…fication


yoerg:
🚀 I'd also suggest changing the test values in _Test/FMI3/parser_test_xmls/type_definitions/valid/clock/modelDescription.xml_ from `2.0` to e.g. `2.1`. That might have uncovered the bug earlier.

PeterMeisrimelModelon:
> 🚀 I'd also suggest changing the test values in _Test/FMI3/parser_test_xmls/type_definitions/valid/clock/modelDescription.xml_ from `2.0` to e.g. `2.1`. That might have uncovered the ~bug~ _inconsitency_ earlier.

Good suggestion; added it.

PR #185, opened 2025-05-07
CMake: Allow running tests when cross-compiling

jschueller:
one must use the add_test(NAME.. COMMAND) version for cmake to use the CROSSCOMPILING_EMULATOR variable if needed


jschueller:
@modelonrobinandersson what about this one ?

Test/test_installation/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ set_property(TARGET fmilib PROPERTY IMPORTED_LOCATION "${FMIL_HOME}/lib/${CMAKE_
 # Build the test
 add_executable(main main.c)
 if(WIN32)
-    target_link_libraries(main PRIVATE fmilib Shlwapi)
+    target_link_libraries(main PRIVATE fmilib shlwapi)
```
modelonrobinandersson, line 56: Can you remind me why was this changed? I felt like I saw this change recently in another PR too but now I am not sure.

Rest of the changes look good!

jschueller: its needed when compiling on a case-sensitive fs like with linux/mingw

PR #186, opened 2025-05-07
Fix some warnings

jschueller:
Just some strncpy warnings where it needs one extra char for the terminating null char


jschueller:
@modelonrobinandersson this one should be simpler to review

PR #190, opened 2025-05-15
Minizip fixes

jschueller:
Small fixes after #169

- Fix build on windows with system minizip
- Fix some format warnings

### DIFF
--- a/src/ZIP/src/fmi_miniunz.c
+++ b/src/ZIP/src/fmi_miniunz.c
@@ -68,8 +68,8 @@
 #define MAXFILENAME (256)
 
 #ifdef _WIN32
-#define USEWIN32IOAPI
-#include "iowin32.h"
+/* MODIFICATION: To allow smoother integration with system minizip-ng, don't use the original minizip's iowin32 API */
+#include <windows.h>
 #endif
 
 /**
--- a/src/ZIP/src/fmi_minizip.c
+++ b/src/ZIP/src/fmi_minizip.c
@@ -66,8 +66,8 @@
 #include "zip.h"
 
 #ifdef _WIN32
-        #define USEWIN32IOAPI
-        #include "iowin32.h"
+/* MODIFICATION: To allow smoother integration with system minizip-ng, don't use the original minizip's iowin32 API */
+#include <windows.h>
 #endif
 
 
--- a/Test/FMI2/fmi2_enum_test.c
+++ b/Test/FMI2/fmi2_enum_test.c
@@ -33,7 +33,7 @@ static int validate_initial(fmi2_causality_enu_t c,
     fmi2_variability_enu_t v = fmi2_variability_enu_constant;
     for (i = 0; i < N_VARIABILITIES; i++, v++) {
         fmi2_initial_enu_t initial = fmi2_get_default_initial(v, c);
-        ASSERT_MSG(initial == expected[i], fmi2_variability_to_string(v));
+        ASSERT_MSG(initial == expected[i], "%s", fmi2_variability_to_string(v));
     }
 
     return CTEST_RETURN_SUCCESS;
@@ -144,7 +144,7 @@ int validate_variability_causality(fmi2_causality_enu_t c, int expected[])
 
     for (i = 0; i < N_VARIABILITIES; i++, v++) {
         int is_valid = fmi2_is_valid_variability_causality(v, c);
-        ASSERT_MSG(is_valid == expected[i], fmi2_variability_to_string(v));
+        ASSERT_MSG(is_valid == expected[i], "%s", fmi2_variability_to_string(v));
     }
 
     return CTEST_RETURN_SUCCESS;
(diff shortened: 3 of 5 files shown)


jschueller:
what do you think @filip-stenstrom ?

src/ZIP/src/fmi_miniunz.c
```diff
@@ -62,13 +62,17 @@
 
 #include "unzip.h"
 
+// undefined with minizip-ng
+#ifndef Z_BZIP2ED
+#define Z_BZIP2ED 12
+#endif
```
filip-stenstrom, line 68: Why is this necessary? From what I can see nothing in this translation unit depends on this define.

src/ZIP/src/fmi_miniunz.c
```diff
@@ -31,7 +31,7 @@
         #endif
 #endif
 
-#if defined(__APPLE__) || defined(__HAIKU__) || defined(MINIZIP_FOPEN_NO_64)
+#if defined(__APPLE__) || defined(__HAIKU__) || defined(MINIZIP_FOPEN_NO_64) || defined(_WIN32)
```
filip-stenstrom, line 34: Are there issues with win32 this tries to solve? Note that _WIN32 is set also for win64, and likely we still want to use fopen64 etc for win64.

https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170&redirectedfrom=MSDN

src/ZIP/src/fmi_miniunz.c
```diff
@@ -62,13 +62,17 @@
...
+#endif
+
 #define CASESENSITIVITY (0)
 #define WRITEBUFFERSIZE (8192)
 #define MAXFILENAME (256)
 
 #ifdef _WIN32
-#define USEWIN32IOAPI
-#include "iowin32.h"
+#include <windows.h>
```
filip-stenstrom, line 75: Also wonder what the purpose of this is.

filip-stenstrom (commented):
I don't understand the issues solved by the fixes. See threads.

src/ZIP/src/fmi_miniunz.c
jschueller: see line 271:
```
fmi_miniunz.c:271:        if (file_info.compression_method==Z_BZIP2ED)
```

src/ZIP/src/fmi_miniunz.c
jschueller: iowin32.h is from the minizip folder and doesnt exist out of that context

src/ZIP/src/fmi_miniunz.c
jschueller: maybe I missed something

src/ZIP/src/fmi_miniunz.c
filip-stenstrom: I see. Do we know if there are any reasons why there was a need for iowin32.h in this code originally? Given that the Github CI actions seem to pass I'm thinking if it's performance related, or just demonstration of the API.

jschueller: I'm not sure what it does but we cannot use it:
```
#        ifdef USEWIN32IOAPI
        fill_win32_filefunc64A(&ffunc);
        uf = unzOpen2_64(zipfilename,&ffunc);
#        else   
        uf = unzOpen64(zipfilename);
#        endif
```

filip-stenstrom: Minizip-ng seems to have a compatibility layer with original minizip, so possibly we could include copy iowin32.h to ZIP/include/FMI/fmi_iowin32.h and include that instead. Shouldn't cause any changes in that case.

jschueller: what if they decide to remove the old api at some point ?
I'd prefer to use only the new apis

filip-stenstrom: Given that just `unzOpen64` seems to work without using special Windows low-level API, and we don't know why the other code was needed in the first place, let's go with this for now and if there are any issues we can revisit then.

jschueller: ok, great!

src/ZIP/src/fmi_minizip.c
```diff
@@ -66,8 +66,7 @@
 #include "zip.h"
 
 #ifdef _WIN32
-        #define USEWIN32IOAPI
-        #include "iowin32.h"
+#include <windows.h>
```
filip-stenstrom, line 69:
```suggestion
/* MODIFICATION: To allow smoother integration with system minizip-ng, don't use the original minizip's iowin32 API */
#include <windows.h>
```

src/ZIP/src/fmi_miniunz.c
```diff
@@ -68,8 +68,7 @@
...
```
filip-stenstrom, line 71:
```suggestion
/* MODIFICATION: To allow smoother integration with system minizip-ng, don't use the original minizip's iowin32 API */
#include <windows.h>
```

src/ZIP/src/fmi_miniunz.c
filip-stenstrom: Added some suggestions with comments about modifications. If you agree and apply we can merge this.

jschueller: done

filip-stenstrom (approved):

PR #196, opened 2025-05-22
Partial revert of #195 due to MSVC link issues

filip-stenstrom:
On master:

```
fmizip.lib(fmi_minizip.obj) : error LNK2019: unresolved external symbol fseeko referenced in function fmi_minizip [<...>\build_msvc_md\compress_test_fmu_zip.vcxproj]
fmizip.lib(fmi_minizip.obj) : error LNK2019: unresolved external symbol ftello referenced in function fmi_minizip [<...>\build_msvc_md\compress_test_fmu_zip.vcxproj]
```

In #195 this didn't appear to be needed for fmi_minizip.c (only fmi_miniunz.c), but was added to keep changes similar. However it appears that didn't work out due to different subset of macro-defined functions being used.

### DIFF
--- a/src/ZIP/src/fmi_minizip.c
+++ b/src/ZIP/src/fmi_minizip.c
@@ -33,8 +33,7 @@
         #endif
 #endif
 
-/* MODIFICATION: fopen64 etc. are not defined for MSVC. Original minizip handles this via ioapi.h, but minizip-ng doesn't. */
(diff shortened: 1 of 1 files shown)


filip-stenstrom:
@jschueller I'm reverting part of #195 since MSVC with bundled Minizip is broken. Given that the reverted part wasn't originally necessary in your PR I hope this won't cause issues.

PeterMeisrimelModelon (approved):

jschueller:
no problem
but in the github CI we do test for msvc + bundled minizip, it should have caught it isnt it ?

PR #199, opened 2025-06-06
Correct printf format strings

tbleher:
Add a macro to make gcc and clang check the format strings of all printf-like functions. Fix any errors found by gcc.

Note: the new macros can only attach to function declarations. Where there was no previous declaration (some static functions), a new function declaration was added directly before the definition.

### DIFF
--- a/Test/FMI3/fmi3_import_bouncingball_xml_test.cpp
+++ b/Test/FMI3/fmi3_import_bouncingball_xml_test.cpp
@@ -27,6 +27,7 @@
 #include <stdarg.h>
 #include <assert.h>
 #include <time.h>
+#include <inttypes.h>
 
 #include "fmilib.h"
 #include "config_test.h"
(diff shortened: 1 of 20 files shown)


Test/FMI3/fmi3_import_bouncingball_xml_test.cpp
```diff
@@ -228,7 +229,7 @@ void printVariableInfo(fmi3_import_t* fmu,
             break;
         }
         case fmi3_base_type_enum:{
-            printf("start = %lld\n", fmi3_import_get_enum_variable_start(fmi3_import_get_variable_as_enum(v)));
+            printf("start = %" PRId64 "\n", fmi3_import_get_enum_variable_start(fmi3_import_get_variable_as_enum(v)));
```
PeterMeisrimelModelon, line 232:
```suggestion
            // doesn't work on mingw_64, but there are enums in bouncing ball
            // printf("start = %" PRId64 "\n", fmi3_import_get_enum_variable_start(fmi3_import_get_variable_as_enum(v)));
```

Not worth fixing just for mingw_64. These tests are largely conversions from fmi1 & fmi2 and were manually tested/inspected. Plus there are no enums in bouncingBall; so this part of the code is not executed either way.

PeterMeisrimelModelon (approved):
Thanks for your contribution!
Just a minor issue popped up when testing with mingw. (hope there not more after)

tbleher:
> Thanks for your contribution! Just a minor issue popped up when testing with mingw. (hope there not more after)

Thanks for reviewing and testing this! I have applied the suggestion.

PR #202, opened 2025-06-23
Minizip: use localtime_r instead of localtime on POSIX systems

tbleher:
`localtime` is not thread-safe on POSIX systems, meaning that the values might be corrupted if another thread in the same process calls `localtime` at the same time. Use `localtime_r` instead, which is thread-safe.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,8 @@ Note that version 2.1 is the first version with release notes. Please see the co
 
 ## FUTURE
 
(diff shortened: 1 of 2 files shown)


modelonrobinandersson:
Thank you @tbleher! Could you add a changelog entry as well?

tbleher:
> @tbleher you are correct, `RELEASE_NOTES.md`, sorry for the confusion!

Thank you for the clarification :+1:  I have added a line there, I hope it is suitable.

PR #210, opened 2025-12-03
CMake: Set soversion

jschueller:


PeterMeisrimelModelon:
Sorry for forgetting about this one for a while; Can you elaborate a bit on the purpose of this?

Issue #209, opened 2025-11-25
Cross-compiled library assumes wrong binaries directory


cbu-hw:
**Describe the bug**
When cross-compiling FMIL for Windows on a Linux Host, an incorrect CMake configuration causes the library to search for FMU binaries in the `linux64` directory instead of `win64`.

**To Reproduce**
Run the following Dockerfile on Windows, with an `output` directory at the same level. This checks out the latest FMIL, fetches reference FMUs, sets up the cross-compilation toolchain, and exports the build folder to `output`.
`docker build --progress=plain -t fmil-test:latest --output output .`

Then, run in the output folder: `.\fmi_import_test.exe .\fmus\2.0\BouncingBall.fmu`

```Dockerfile
# Troubleshoot cross-compiling fmi-library
FROM ubuntu:24.04 AS build

RUN apt update && export DEBIAN_FRONTEND=noninteractive \
&& apt -y install --no-install-recommends wget git ninja-build cmake g++-mingw-w64 zip unzip \
&& apt clean all \
&& rm -rf /var/lib/apt/lists

# Check out b711f5d
RUN git clone --depth 1 --branch master https://github.com/modelon-community/fmi-library.git  \
	&& cd fmi-library \
	&& git reset --hard b711f5d
WORKDIR /fmi-library

RUN wget https://github.com/modelica/Reference-FMUs/releases/download/v0.0.39/Reference-FMUs-0.0.39.zip
RUN unzip Reference-FMUs-0.0.39.zip -d ./fmus

RUN <<HEREDOC cat > mingw-w64-x86_64.cmake
	set(CMAKE_SYSTEM_NAME Windows)
	set(TOOLCHAIN_PREFIX x86_64-w64-mingw32)
	set(CMAKE_SYSTEM_PROCESSOR x86_64)
	# cross compilers to use for C, C++
	set(CMAKE_C_COMPILER \${TOOLCHAIN_PREFIX}-gcc)
	set(CMAKE_CXX_COMPILER \${TOOLCHAIN_PREFIX}-g++)
	set(CMAKE_RC_COMPILER \${TOOLCHAIN_PREFIX}-windres)
	# target environment on the build host system
	set(CMAKE_FIND_ROOT_PATH /usr/\${TOOLCHAIN_PREFIX})
	# modify default behavior of FIND_XXX() commands
	# search for programs in the build host environment
	set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
[...]

PR #213, opened 2026-01-20
fmiplatform: Use target platform

jschueller:
see #209


jschueller:
hi @PeterMeisrimelModelon, any thoughts ?

PR #218, opened 2026-06-10
FMI parsing: accept patch-level FMI 3.0.x versions

### LINKED ISSUE #217: FMI Version Parsing Is Too Strict for FMI 3.0.x FMUs
Description

The FMI version parsing logic currently requires an exact match of the fmiVersion attribute value. For FMI 3.0, the parser accepts only:

else if( strcmp(fmiVersion, "3.0") == 0 ) {
    ...
}

As a result, FMUs whose modelDescription.xml contains:

<fmiModelDescription
    fmiVersion="3.0.1"
    ...
>

are rejected with the error:

XML specifies FMI standard version unknown

Motivation

Although the FMI 3.0 specification states that the value of the fmiVersion attribute should be "3.0" and that future minor revisions will be denoted as "3.1", "3.2", etc., some FMU generators have started producing patch-level version strings such as "3.0.1".

From an interoperability perspective, these FMUs are typically intended to be FMI 3.0-compliant and could be imported successfully if the parser handled patch-level version numbers more leniently.

Proposed Improvement

Consider accepting patch-level versions that are compatible with FMI 3.0 and mapping them to the existing FMI 3.0 implementation.

For example:

3.0
3.0.1
3.0.2
3.0.x

could all be interpreted as FMI 3.0.

This could be implemented by parsing the version components instead of performing an exact string comparison, or by accepting strings that start with "3.0" and are followed by a valid patch suffix.

Benefits

Improved robustness when importing FMUs generated by third-party tools.
Better interoperability across the FMI ecosystem.
Avoids rejecting FMUs that are effectively FMI 3.0-compatible due to a formatting difference in the version string.

Expected Behavior

An FMU declaring:

<fmiModelDescription
    fmiVersion="3.0.1"
    ...
>

should be recognized as an FMI 3.0 FMU rather than being rejected as an unknown FMI version.

Additional Note

This request is not intended to change the interpretation of FMI minor versions (e.g. 3.1, 3.2), which may introduce new functionality and should continue to be handled explicitly. The goal is only to make the parser tolerant of patch-level version identifiers that remain compatible with FMI 3.0.

jschueller:
Replace exact strcmp("3.0") with a prefix check that also accepts "3.0.1", "3.0.x", etc. as FMI 3.0, while rejecting "3.1", "3.2".

Closes #217

### DIFF
--- a/src/XML/src/FMI/fmi_xml_context.c
+++ b/src/XML/src/FMI/fmi_xml_context.c
@@ -105,7 +105,7 @@ void XMLCALL fmi_xml_parse_element_start(void *c, const char *elm, const char **
         XML_StopParser(context->parser,0);
         return;
     }
-    else if( strcmp(fmiVersion, "3.0") == 0 ) {
+    else if( strncmp(fmiVersion, "3.0", 3) == 0 && (fmiVersion[3] == '\0' || fmiVersion[3] == '.') ) {
         jm_log_verbose(context->callbacks, MODULE, "XML specifies FMI 3.0");
         context->fmi_version = fmi_version_3_0_enu;
         XML_StopParser(context->parser,0);


PeterMeisrimelModelon:
Thanks for the PR; I took the proposed fix and also ported it to the other versions + added a unit test for it, see https://github.com/modelon-community/fmi-library/pull/219

PR #219, opened 2026-06-24
Adding support for parsing patch releases for modelDescription fmiVersion

PeterMeisrimelModelon:
Building on top of https://github.com/modelon-community/fmi-library/pull/218; copying the fix to all FMI versions  + adding test & release note.

There is no real API for just fetching the fmiVersion from a modelDescription.xml, hence I duplicated the dummy FMU test.


src/XML/src/FMI/fmi_xml_context.c
```diff
@@ -93,19 +93,19 @@ void XMLCALL fmi_xml_parse_element_start(void *c, const char *elm, const char **
...
     }
-    else if( strcmp(fmiVersion, "2.0") == 0 ) {
+    else if( strcmp(fmiVersion, "2.0") == 0 && (fmiVersion[3] == '\0' || fmiVersion[3] == '.') ) {
         jm_log_verbose(context->callbacks, MODULE, "XML specifies FMI 2.0");
         context->fmi_version = fmi_version_2_0_enu;
         XML_StopParser(context->parser,0);
         return;
     }
-    else if( strcmp(fmiVersion, "3.0") == 0 ) {
+    else if( strncmp(fmiVersion, "3.0", 3) == 0 && (fmiVersion[3] == '\0' || fmiVersion[3] == '.') ) {
```
efredriksson-modelon: Should we not use `strncmp` also for the 1.0 and 2.0 checks? Best would probably be to break out a function that does this looser check of version number.

Also, this means that versions like `3.0.anything` will be accepted. It is probably fine but is a bit strange.

efredriksson-modelon (changes requested):

src/XML/src/FMI/fmi_xml_context.c
PeterMeisrimelModelon: Right, good catch. 

I broke it out into a separate helper function that does a proper processing and splits into major,minor & patch. With this we should be safe for the future.

src/XML/src/FMI/fmi_xml_context.c
```diff
@@ -73,9 +75,68 @@ void fmi_xml_fatal(fmi_xml_context_t *context, const char* fmt, ...) {
     XML_StopParser(context->parser,0);
 }
 
+/* Returns non-zero if 'str[0..len)' is empty or contains any non-digit character. */
```
efredriksson-modelon: `Returns non-zero` -> `Returns zero`?

efredriksson-modelon (commented):
Looks good 🧇 

I ran an AI review and it found one thing that is interesting. The terminals code will not currently handle patch versions and will log and drop them:
```
const char* termIconfmiVersion = jm_vector_char2string(&(termIcon->fmi3_xml_standard_version));
        if (strcmp(modelDescrfmiVersion, termIconfmiVersion)) {
            fmi3_xml_parse_fatal(context,
                                 "Mismatch of attribute 'fmiVersion' in modelDescription.xml: '%s' and terminalsAndIcons.xml: '%s'.", 
                                 modelDescrfmiVersion,
                                 termIconfmiVersion);
            return -1;
        }
```

Not sure if we want to address that in this PR or a separate one. Or if this is acceptable.

PeterMeisrimelModelon:
> Looks good 🧇
> 
> I ran an AI review and it found one thing that is interesting. The terminals code will not currently handle patch versions and will log and drop them:
> 
> ```
> const char* termIconfmiVersion = jm_vector_char2string(&(termIcon->fmi3_xml_standard_version));
>         if (strcmp(modelDescrfmiVersion, termIconfmiVersion)) {
>             fmi3_xml_parse_fatal(context,
>                                  "Mismatch of attribute 'fmiVersion' in modelDescription.xml: '%s' and terminalsAndIcons.xml: '%s'.", 
>                                  modelDescrfmiVersion,
>                                  termIconfmiVersion);
>             return -1;
>         }
> ```
> 
> Not sure if we want to address that in this PR or a separate one. Or if this is acceptable.

Good catch, I had not considered this interaction at all.

There is however currently no conflict here at all. The affected part of the code is for when we "peek" into the `modelDescription.xml` to determine which principle `fmiVersion` we are dealing with.

The terminalsAndIcons part happens after we have parsed the `modelDescription.xml` and have read in the raw `fmiVersion` strings both and check that they agree exactly. I.e.,

* modelDescription.xml -> fmiVersion = 3.0.1 & terminalsAndIcons.xml -> fmiVersion = 3.0.1 => match, all OK
* modelDescription.xml -> fmiVersion = 3.0.2 & terminalsAndIcons.xml -> fmiVersion = 3.0.1 => mismatch, fails when reading terminalsAndIcons.xml, due to the mismatch. This might be a bit strict, but it also points to some structural mismatch in the FMU. I am OK with starting off strict here; we can loosen this in the future.
* modelDescription.xml -> fmiVersion = 3.0.2a & terminalsAndIcons.xml -> fmiVersion = 3.0.2a => invalid format for modelDescription.xml -> fmiVersion rejected by the peeking

efredriksson-modelon (approved):

PR #220, opened 2026-07-20
fmi3_xml_parser: Allow INF values

### LINKED ISSUE #207: ±∞ start values in FMI 3.0 model description are rejected
If a FMI 3.0 FMU defines a parameter/input start value of `INF` or `-INF`, e.g.
````
<Float64 name="realInput" valueReference="1" causality="input" variability="continuous" start="INF" />
````

`fmi3_import_parse_xml` results in a parser error:
````
Error parsing XML: FMI3XML: min/max value boundary check for parsing Float64 failed
FMI3XML: XML element 'Float64': failed to parse attribute start='INF'
````

The same happens for `Float32` variables with infinite start values.

This [looks to be deliberate](https://github.com/modelon-community/fmi-library/blob/bb232e8b1cc44a590cd2d460dbf6dc3590fe0a69/src/XML/src/FMI3/fmi3_xml_parser.c#L642), but in my reading of [Table 15. Type element and attribute details](https://fmi-standard.org/docs/3.0.2/#min), infinite floating point start values should be allowed on IEEE 754 platforms:

> If not defined, the minimum is the largest negative number that can be represented on the machine for that `<TypeDefinition>`.

The FMI **_2.0_** XML parser accepts infinite start values.

### PR DESCRIPTION
**Pull request by jschueller**
Closes #207

### CODE REVIEW

File: src/XML/src/FMI3/fmi3_xml_parser.c [hunk 1]
```diff
@@ -634,6 +640,19 @@ static int fmi3_xml_str_to_floatXX(fmi3_xml_parser_context_t* context, int requi
         value = &valReadBuff;
     }
 
+    if (isInf) {
```
aicherst: I would prefer to use `isinf` from `math.h` instead of the manual string comparison, because that is case sensitive and would break for e.g. `Inf` or `inf`.
```suggestion
    if (isinf(valReadBuff)) {
```

src/XML/src/FMI3/fmi3_xml_parser.c
aicherst: The comment here could be removed.

aicherst (commented):
Thank you for the PR. We had the same issue in a project at BMW and this fixes the problem.

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -634,11 +644,22 @@ static int fmi3_xml_str_to_floatXX(fmi3_xml_parser_context_t* context, int requi
...
+        switch(primType->bitness) {
+        case fmi3_bitness_64:
+            *(fmi3_float64_t*)field = (fmi3_float64_t)valReadBuff; break;
+        case fmi3_bitness_32:
+            *(fmi3_float32_t*)field = (fmi3_float32_t)valReadBuff; break;
+        default:
+            assert(0); /* impl. error */
+        }
+        return 0;
+    }
```
aicherst: The issue was the `useDefault = 1` path, because in that case `valReadBuff` was never set. The suggested version should correctly detect `inf` for both parsed and default values. Sorry for the confusion.

```suggestion
    if (isinf(*(fmi3_float_buf_t*)value)) {
        /* INF values are valid per FMI spec; assign directly without boundary check */
        switch(primType->bitness) {
        case fmi3_bitness_64:
            *(fmi3_float64_t*)field = (fmi3_float64_t)*(fmi3_float_buf_t*)value; break;
        case fmi3_bitness_32:
            *(fmi3_float32_t*)field = (fmi3_float32_t)*(fmi3_float_buf_t*)value; break;
        default:
            assert(0); /* impl. error */
        }
        return 0;
    }
```

aicherst (approved):

aicherst:
@PeterMeisrimelModelon This is an issue we had in a project at BMW and we would like to fix it properly upstream. Could you please have a look at it?

src/XML/src/FMI3/fmi3_xml_parser.c
```diff
@@ -634,11 +635,22 @@ static int fmi3_xml_str_to_floatXX(fmi3_xml_parser_context_t* context, int requi
         value = &valReadBuff;
     }
 
+    if (isinf(*(fmi3_float_buf_t*)value)) {
```
tbleher, line 649: This causes a (IMHO correct) compiler warning (at least when I cherry-pick on master - haven't tested the original branch):

```
In file included from src/XML/src/FMI3/fmi3_xml_parser.c:19:
src/XML/src/FMI3/fmi3_xml_parser.c: In function ‘fmi3_xml_parse_attr_as_float32’:
src/XML/src/FMI3/fmi3_xml_parser.c:638:15: warning: array subscript ‘fmi3_float_buf_t {aka double}[0]’ is partly outside array bounds of ‘fmi3_float32_t[1]’ {aka ‘float[1]’} [-Warray-bounds]
  638 |     if (isinf(*(fmi3_float_buf_t*)value)) {
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~
src/XML/src/FMI3/fmi3_xml_parser.c:699:67: note: while referencing ‘defaultVal’
  699 |             int required, fmi3_##TYPE##_t* field, fmi3_##TYPE##_t defaultVal) {                                            \
      |                                                   ~~~~~~~~~~~~~~~~^~~~~~~~~~
src/XML/src/FMI3/fmi3_xml_parser.c:702:1: note: in expansion of macro ‘gen_fmi3_xml_parse_attr_as_TYPEXX’
  702 | gen_fmi3_xml_parse_attr_as_TYPEXX(float32, floatXX)
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

One way to fix this would be using the patch below:
```diff
--- a/src/XML/src/FMI3/fmi3_xml_parser.c
+++ b/src/XML/src/FMI3/fmi3_xml_parser.c
@@ -626,7 +626,15 @@ static int fmi3_xml_str_to_floatXX(fmi3_xml_parser_context_t* context, int requi
 
     /* get the value */
     if (!strVal && !required) {
-        value = defaultVal;
+        /* Copy the default value into the 64 bit buffer, so that 'value' always points to a
+         * fmi3_float_buf_t. Otherwise reading it as fmi3_float_buf_t below would be out-of-bounds
+         * when the caller's default is a 32 bit float. */
+        if (primType->bitness == fmi3_bitness_64) {
+            valReadBuff = (fmi3_float_buf_t)(*(fmi3_float64_t*)defaultVal);
+        } else {
+            valReadBuff = (fmi3_float_buf_t)(*(fmi3_float32_t*)defaultVal);
+        }
+        value = &valReadBuff;
         useDefault = 1;
     } else {
         if (sscanf(strVal, formatter, &valReadBuff) != 1) {
```

jschueller: ok, included

